PR #1, opened 2020-02-10
refactor: move config to configs service

austin-king:

Just moving config values into a config service to manage our state in a centralized location. @hbergren I copied over some of the linting config (in the initial commit, not this PR). Would prob be worth glancing over to see if there are any issues with it.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,17 +1,17 @@
+import Config from './services/config'
+
 import Express = require('express')
+import Reduct = require('reduct')
 
 export default class App {
   private publicApiServer: Express.Application
 
   private privateApiServer: Express.Application
 
-  private publicApiServerPort: number
+  private config: Config
 
-  private privateApiServerPort: number
-
-  constructor() {
-    this.publicApiServerPort = 8080
-    this.privateApiServerPort = 8081
+  constructor(deps: Reduct.Injector) {
+    this.config = deps(Config)
   }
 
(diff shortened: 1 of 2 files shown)


src/app.ts
```diff
@@ -1,17 +1,17 @@
+import Config from './services/config'
+
 import Express = require('express')
+import Reduct = require('reduct')
```
hbergren: What is `Reduct`?

src/services/config.ts
```diff
@@ -0,0 +1,10 @@
+export default class Config {
+  public privateApiPort: number
+
+  public publicApiPort: number
+
+  constructor() {
+    this.publicApiPort = 8080
+    this.privateApiPort = 8081
+  }
```
hbergren: You might be able to use class fields to define the numbers right above, like:

```js
class Config {
 publicApiPort = 8080
 privateApiPort = 8081
}
```

If not, could you switch the order of the class members to match the order in the constructor?

src/app.ts
austin-king: It's a functional dependency manager, I basically use it to force myself to keep my services stateless and manage all that in a config file. It's just a tool I've gotten used to using, happy to remove if you think would be confusing to other maintainers. Just lmk.

src/services/config.ts
austin-king: Good point, pushed fix.

PR #2, opened 2020-02-12
add routes

austin-king:

Routing based upon paths with functionality stubbed out.


src/services/users.ts
```diff
@@ -0,0 +1,39 @@
...
+      } else if (req.type === 'delete_user') {
+        await this.deleteUser(req, res, next)
+      }
+    } catch (e) {
+      console.error(e)
+      // TODO: add in error reporting for failed requests
+    }
+
+    public async getUser(req: Request, res: Response, next: NextFunction): Promise<void> {
+      // TODO: add DB request
```
keefertaylor: I tend to like throwing errors rather than silently doing nothing. 

```
throw new Error("unimplemented")
```

This will make it easy to debug what happened if you hit one of these later.

src/services/config.ts
```diff
@@ -0,0 +1,10 @@
+export default class Config {
+  public privateApiPort: number
+
+  public publicApiPort: number
```
keefertaylor: Also would encourage you to doc this class.

src/services/users.ts
```diff
@@ -0,0 +1,39 @@
+import { Request, Response, NextFunction } from 'express'
+
+export default class Users {
+  constructor() {
+  }
+
+  public async middleware(req: Request, res: Response, next: NextFunction): Promise<void> {
+    try {
+      if (req.type === 'get_user') {
```
keefertaylor: Would this be a good case for an enum?

src/services/users.ts
```diff
@@ -0,0 +1,39 @@
...
+        await this.getUser(req, res, next)
+      } else if (req.type === 'post_user') {
+        await this.postUser(req, res, next)
+      } else if (req.type === 'put_user') {
+        await this.putUser(req, res, next)
+      } else if (req.type === 'delete_user') {
+        await this.deleteUser(req, res, next)
+      }
+    } catch (e) {
+      console.error(e)
```
keefertaylor: Does this build? I think you need to end the function after this catch block. 

If this is some random javascript I don't understand, I think you should at least return in the catch block.

src/services/config.ts
```diff
@@ -0,0 +1,10 @@
+export default class Config {
+  public privateApiPort: number
```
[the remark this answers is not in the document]
austin-king: happy to standardize

src/services/users.ts
austin-king: good practice

keefertaylor:
Is there a design doc for this feature?

PR #5, opened 2020-02-24
feat: standardize linting according to wallet spec

austin-king:

Standardizes PayID to use our linting scheme used elsewhere. Uses the updated version of .eslintrc.js in the wallet

### Type of Change


.eslintrc.js
```diff
@@ -50,23 +50,42 @@ module.exports = {
...
+      },
+    ],
+
+    // TODO: @hbergren come back and delete all these disables
+    // These are rules that were violated before ESLint was fully implemented and we haven't had time to fix yet
+    'no-console': 0,
+    'no-shadow': 0,
+    'import/no-cycle': 0,
+    'import/no-unresolved': 0,
+    'max-classes-per-file': 0,
```
Stormtv: If we can we should not have these in this library. These are overrides that shouldn't exist but did exist in the Xpring-Wallet-System due to implementing new linting standards.

.eslintrc.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
   parserOptions: {
     // Enable linting rules with type information from our tsconfig
     tsconfigRootDir: __dirname,
-    project: ['./tsconfig.eslint.json'],
+    project: ['./tsconfig.json'],
```
Stormtv: I'm a bit out of context here, what is the reasoning behind this change?

.eslintrc.js
austin-king: @hbergren should I just delete these right now? Or should I leave them until they are deleted out of xpring-wallet as well?

.eslintrc.js
austin-king: PayID is a new repo and I was asked to standardize linting according to our other repos, so I just took the exact linting spec from the wallet

Stormtv: Ah okay, in that case we should be able to delete the `./tsconfig.eslint.json` file from this repository. In Xpring Wallet System we don't have tests for the server, if we are planning on having a test directory in the future for this project we should avoid making this change.

austin-king: This repo will have tests in the future

hbergren: If this repo will have tests, I would keep this `tsconfig.eslint.json` and copy over the corresponding `tsconfig.eslint.json` file from whatever repo you cloned this all from.

That is necessary to keep ESLint working with type information from `tsconfig` while not breaking when you add a `/tests` directory.

.eslintrc.js
hbergren: I would delete them now, unless you already have violations of these rules that you don't want to fix.

.eslintrc.js
austin-king: Cool, I originally copied .eslintrc.js from `xpring-wallet-system` but that no longer has a `tsconfig.eslint.json` in it so I took the one from xpring-bounty-bot.

.eslintrc.js
austin-king: Cool, deleted.

.eslintignore
```diff
@@ -0,0 +1,7 @@
+# Do not lint generated files.
+src/generated
```
Stormtv: This line shouldn't be needed this is for the generated protobuff files

PR #6, opened 2020-02-24
feat: private API skeleton

austin-king:

PayID is a server with two types of endpoints:
public -> responds to $payment.pointers. This will be a simple mapping of $payment.pointer to crypto address based upon the incoming content-type header. Today it is just a dumb express server.

private -> this will be how partners can post/get/put/delete users to our database. For example, if cryptowallet.com wants to have $payment.pointers but not host a server themselves they will redirect a subdomain (ex: `pay.cryptowallet.com`) to our service and then they will post users along with their addresses to our service. 
ex: if they POST a new user alice and her crypto address we will be able to resolve `pay.cryptowallet.com/alice` to her address anytime people query `$pay.cryptowallet.com/alice`.
Today the private API is just a dumb express server + a route based skeleton. All functionality is stubbed out today so it just logs which HTTP method was received. 

@keefertaylor I refactored pretty heavily from the previous PR so I just closed that and opened this one in hopes that it would be cleaner and easier to comprehend. I believe all of the concerns mentioned in the last PR should have changes reflected here. 

### Context of Change

Here is the design doc for the API interface if you're looking for greater context: https://docs.google.com/document/d/1xpFO9qH9wzQvRTY426VEDIEktqmOunmqZVldPhHnwWI/edit

-->

### Type of Change

## Before / After

Before: 2 simple express servers (private & public)
After: 1 simple express server (public) and 1 simple express server with a routes skeleton (private)

## Test Plan

Locally ran `curl -X {GET || POST || PUT || DELETE} localhost:8081/v1/users` to ensure the log matched the HTTP method and the response matched the HTTP method. As we expand the functionality past just stubs we'll have this in a testing suite to ensure that each HTTP method performs the desired action.

NOTE: if you pull and lint locally you will see errors because I am still waiting for my other PR w/ updating linting spec to be merged. I confirmed it works w/ the updated spec by checking out the other branch's (https://github.com/xpring-eng/payid-api/pull/5) linting files.


src/services/api.ts
```diff
@@ -0,0 +1,18 @@
+import { Response, NextFunction } from 'express'
+
+import { PayIDRequest } from '../types/payid-request'
+
+export default class API {
+  static successMiddleware(statusCode: number) {
```
keefertaylor: I'm sure this is wildly obvious to you but for those of us who haven't been writing express apps for O(years), could you just doc what these are doing? 

Presumably this is middleware that sets httpCode = 200. I'm not quire sure what the `type` middleware is doing.

---
Optional: It also strikes me that maybe these could use better variable names and that could help me understand.  Maybe `setStatusToSuccessMiddleware` and ??. If this is idiomatic naming and I'm just an express noob feel free to ignore.

src/services/config.ts
```diff
@@ -1,5 +1,19 @@
 export default class Config {
-  public publicApiPort = 8080
+  /**
+   * @param publicAPIPort port responding to PayID requests
+   * @param privateAPIPort port responding to API updating users' PayIDs
+   * @param version versioning enabling future upgrades
```
keefertaylor: Would version be better as an enum rather than an untyped string?

src/services/users.ts
```diff
@@ -0,0 +1,72 @@
+import { Response, NextFunction } from 'express'
+
+import { PayIDRequest } from '../types/payid-request'
+
+export default class Users {
+  /**
+   * @param methods all possible methods for our user interface
+   */
+  static readonly methods = {
```
keefertaylor: Would this be better as an enum type?

src/services/users.ts
```diff
@@ -0,0 +1,72 @@
...
+        Users.postUser(req, res, next)
+        break
+      case Users.methods.PUT_USER:
+        Users.putUser(req, res, next)
+        break
+      case Users.methods.DELETE_USER:
+        Users.deleteUser(req, res, next)
+        break
+      default:
+        throw new Error(
```
keefertaylor: Won't this just crash the server process? I feel like we should throw back a 400 / bad request here.

src/services/users.ts
```diff
@@ -0,0 +1,72 @@
...
+          `Request of type ${req.type} does not match existing user methods.`,
+        )
+    }
+  }
+
+  static getUser(_req: PayIDRequest, res: Response, _next: NextFunction): void {
+    res.send('TODO: get user')
+    console.log('get user hit')
+    // TODO: add DB _request
```
keefertaylor: We've adopted named TODOs on other repos and I'd ask you to consider doing the same. Generally, a TODO should have an owner and a clear AI attached to it. If the AI is bigger than a line or hard to describe, it's probably better as an asana task. 

```
TODO(aking): Implement fetching from a database
```

Attaching names to things makes it clear who has context. I get that it's contrived here but it's a good habit.

keefertaylor: + other TODOs in this PR

src/services/privateAPI.ts
```diff
@@ -0,0 +1,53 @@
...
+    this.users = deps(Users)
+  }
+
+  public init(server: Application): void {
+    server.get(
+      Config.version + this.userPath,
+      API.typeMiddleware(Users.methods.GET_USER),
+      Users.middleware.bind(this.users),
+      API.successMiddleware(200),
+    )
```
hbergren: In plain English, what does this `server.get()` call do (with all the parameters explained).

(I'm unfamiliar with Express).

src/services/config.ts
```diff
@@ -1,5 +1,19 @@
...
+   */
+  static readonly publicAPIPort = 8080
 
-  public privateApiPort = 8081
+  static readonly privateAPIPort = 8081
+
+  static readonly version = '/v1'
```
hbergren: Good job thinking about API versioning!

src/app.ts
```diff
@@ -1,45 +1,38 @@
...
 
-  private privateApiServer: Express.Application
+  private privateAPIServer: Express.Application
 
-  private config: Config
+  private readonly privateAPI: PrivateAPI
 
   constructor(deps: Reduct.Injector) {
-    this.config = deps(Config)
+    this.privateAPI = deps(PrivateAPI)
```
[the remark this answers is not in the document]
austin-king: Instantiating the class in a way that forces you to only have one instance. Personal forcing function to ensure I don't leave state laying around in places it shouldn't be. Hans pointed this out earlier as well, Reduct is just a package I'm used to using but if you think this would be confusing to other devs on the team or other devs in general (b/c this is open source) happy to replace it for normal instantiation.

src/app.ts
```diff
@@ -1,45 +1,38 @@
...
+    this.publicAPIServer = Express()
+    this.publicAPIServer.get('/', (req, res) => {
       res.send('This server should receive requests from $domain.com/user')
     })
-    this.privateApiServer.get('/', (req, res) => {
-      res.send(
-        'This will expose a simple CRUD API for wallets to update user mappings.',
-      )
-    })
+    this.publicAPIServer.listen(Config.publicAPIPort)
```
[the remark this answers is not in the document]
austin-king: Fixed

PR #7, opened 2020-02-25
Add CircleCI Configuration

keefertaylor:

Create a circle CI configuration. 

### Context of Change

We need to drop this config into the repo before we can enable Circle to build. This also sets up CodeCov. 

It's likely that this may fail on first run and we'll need a follow up PR. 

### Type of Change

## Before / After
Tests run

## Test Plan

yolo


keefertaylor:
If you look in the CI it uploaded code coverage. I believe CodeCov won't start commenting on PR reviews until it has a baseline upload from master (which requires this PR to get merged).

.circleci/config.yml
```diff
@@ -0,0 +1,37 @@
+# Copyright Ripple Labs, 2019
+version: 2.1
+orbs:
+  codecov: codecov/codecov@1.0.4
+jobs:
+  build:
+    docker:
+      - image: circleci/node:10.4.0
```
hbergren: node 12? Or are we explicitly supporting Node 10?

PR #534, opened 2020-06-18
Fix broken `xpring-eng` links

hbergren:

Fix various links in documentation that point to `xpring-eng` when they should point to `payid-org`.

### Context of Change

We migrated the repo, and now we need to update links.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

I clicked various links, to make sure they work.


hbergren:
Also removing our Security issue template, because the fact that we have a security policy makes something show up there.

security

PR #538, opened 2020-06-19
Custom PostgreSql port via DB_PORT environment variable

mouradski:
### Context of Change
Some database providers allocate different port per client

### Type of Change
New feature


hbergren:
Thanks for the contribution! Does it make sense to update the [`docker-compose.yml`](https://github.com/payid-org/payid/blob/master/docker-compose.yml#L43) file as well?

mouradski:
i don't think we need to change docker-compose.yml because we use a postgre docker container with default port

hbergren:
> If we're adding it to the config then we should also plug it into our DB interface in src/db/knex.ts I think

That will happen automatically.

```ts
const knexConfig = {
  client: 'pg',
  connection: config.database.connection,
  ...
```

hbergren:
> i don't think we need to change docker-compose.yml because we use a postgre docker container with default port

Well, if you set the port using an ENV var that isn't `5432`, then you wouldn't be able to connect to the Postgres Docker container, right?

mouradski:
@hbergren DB_PORT variable is not present in docker-compose.yml so I don't see how we can use a port different of 5432

If we execute docker-compose up command all should works fine even if a different DB_PORT value is set outside the docker-compose.yml

docker-compose.yml
```diff
@@ -16,6 +16,8 @@ services:
     depends_on:
       - 'db'
     environment:
+      # Force the default PostgreSql port and ignore the value of the host environment variable.
+      - DB_PORT=5432
```
hbergren, line 20: Hey @mouradski , did this work for you?

mouradski: yes, here is my test on ubuntu host : 

```
export DB_PORT = 9999
docker build . -t payid:1.0.3  (for test)
changed tag in docker-compose.yml to 1.0.3

docker-compose up 

```
Output : 

```
payid-database |
payid-database | PostgreSQL Database directory appears to contain a database; Skipping initialization
payid-database |
payid-database | 2020-06-23 14:17:16.575 UTC [1] LOG:  starting PostgreSQL 12.3 on x86_64-pc-linux-musl, compiled by gcc (Alpine 9.3.0) 9.3.0, 64-bit
payid-database | 2020-06-23 14:17:16.575 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
payid-database | 2020-06-23 14:17:16.575 UTC [1] LOG:  listening on IPv6 address "::", port 5432
payid-database | 2020-06-23 14:17:16.575 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
payid-database | 2020-06-23 14:17:16.582 UTC [21] LOG:  database system was interrupted; last known up at 2020-06-23 14:14:31 UTC
payid-database | 2020-06-23 14:17:16.597 UTC [21] LOG:  database system was not properly shut down; automatic recovery in progress
payid-database | 2020-06-23 14:17:16.598 UTC [21] LOG:  redo starts at 0/16B47A0
payid-database | 2020-06-23 14:17:16.599 UTC [21] LOG:  invalid record length at 0/16DDFC0: wanted 24, got 0
payid-database | 2020-06-23 14:17:16.599 UTC [21] LOG:  redo done at 0/16DDF80
payid-database | 2020-06-23 14:17:16.604 UTC [1] LOG:  database system is ready to accept connections
payid    | Postgres is up - executing command
payid    | WARN  PAYID_ORG must be set for metrics to be pushed. Metrics will not be pushed.
payid    | INFO  Public API listening on 8080
payid    | INFO  Private API listening on 8081

```


**but, even if i don't force DB_PORT to 5432 in docker-compose.yml it still works**

mouradski: Apologies, just fixed that, I was pushing directly from github so some tests was not executed :(

/builds/payid-org/payid/src/config.ts
   15:13  error  Unnecessary conditional, expected left-hand side of `??` operator to be possibly null or undefined  @typescript-eslint/no-unnecessary-condition

mouradski: > Hey @mouradski , did this work for you?

Hi @hbergren it's ok for you ? 

I still believe it is not necessary to add the port in the docker-compose.yml, but if you had a problem that I did not see, it must be that there is a reason and we can adding it to avoid possible problems

hbergren: It might not be necessary to add the port in the `docker-compose.yml` file for `npm run devEnvUp` or `docker-compose up`, because Docker is basically just ignoring your local `DB_PORT` environment variable.

But, we also support starting up just the PayID database using `npm run devDbUp` or `docker-compose up -d db`.

If you did that, then right now, the database would boot up on port `5432`, but if you spun up the server using `npm run start`, it would try to connect to the database on port `5433` and fail to connect.

That's why I suggested modifying the `docker-compose.yml` file to template in your local `DB_PORT` environment variable (perhaps with a default value of `5432`), so that `npm run devDbUp` would still work.

Does that explanation make sense?

mouradski: got it :) 

we can use ${DB_PORT} but we need to set a .env file containing default DB_PORT value, but I saw that .env file is present in .gitignore so i'm not sur that is the best solution

mouradski: got it :)

mouradski: > It might not be necessary to add the port in the `docker-compose.yml` file for `npm run devEnvUp` or `docker-compose up`, because Docker is basically just ignoring your local `DB_PORT` environment variable.
> 
> But, we also support starting up just the PayID database using `npm run devDbUp` or `docker-compose up -d db`.
> 
> If you did that, then right now, the database would boot up on port `5432`, but if you spun up the server using `npm run start`, it would try to connect to the database on port `5433` and fail to connect.
> 
> That's why I suggested modifying the `docker-compose.yml` file to template in your local `DB_PORT` environment variable (perhaps with a default value of `5432`), so that `npm run devDbUp` would still work.
> 
> Does that explanation make sense?

got it :)

Tests : 

`docker-compose up` (with/without setting localy DB_PORT)
`docker-compose up  -d db & npm run start`   (with/without setting localy DB_PORT)

hbergren: Hey @mouradski, I still can't get this to work, here's what I have (my shell is fish):

```fish
set -Ux DB_PORT 5433

# Not sure these steps are actually necessary, but I ran them
docker build . -t payid:1.0.3 # (for test)
# changed tag in docker-compose.yml to 1.0.3

npm run devDbUp
docker ps
4894abcc3ce3        postgres:12-alpine   "docker-entrypoint.s…"   5 minutes ago       Up 5 minutes        0.0.0.0:5432->5432/tcp   payid-database
```

Notice how it still says the database of the Docker container is running on `5432` even though my `DB_PORT` environment variable says it should be `5433`.

mouradski: > Hey @mouradski, I still can't get this to work, here's what I have (my shell is fish):
> 
> ```fish
> set -Ux DB_PORT 5433
> 
> # Not sure these steps are actually necessary, but I ran them
> docker build . -t payid:1.0.3 # (for test)
> # changed tag in docker-compose.yml to 1.0.3
> 
> npm run devDbUp
> docker ps
> 4894abcc3ce3        postgres:12-alpine   "docker-entrypoint.s…"   5 minutes ago       Up 5 minutes        0.0.0.0:5432->5432/tcp   payid-database
> ```
> 
> Notice how it still says the database of the Docker container is running on `5432` even though my `DB_PORT` environment variable says it should be `5433`.

@hbergren 
here is my test (it's a ssh session, on ubuntu server, not gitbash)

https://www.youtube.com/watch?v=4BJKJ-S9_j8

all works fine for me, you fetch last commits ?

hbergren: Things definitely work in your video haha. Thanks for taking the screencast.

I've definitely been using your latest code. I'm on Mac OS X and my shell is `fish`. I wonder if that is somehow affecting things?

I'll ask one of my coworkers with more Docker experience to take a look at this.

mouradski: > Things definitely work in your video haha. Thanks for taking the screencast.
> 
> I've definitely been using your latest code. I'm on Mac OS X and my shell is `fish`. I wonder if that is somehow affecting things?
> 
> I'll ask one of my coworkers with more Docker experience to take a look at this.

Thank you for the time you have given me, I appreciate it, you are doing a huge work !!

FlorentSG:
Hi @mouradski , thank you for your contribution.

As @hbergren mentioned, we can run the database container independently using `npm run devDbUp`.
So the change made in `docker-compose.yml` for the `db` service port mapping is necessary.

Nevertheless, there is no need to change the `payid` service by adding:
```
# Local DB_PORT environment variable is used if set (default 5432)
- DB_PORT='${DB_PORT:-5432}'
```
First, I think there is a small typo mistake, for me, it works by removing the quotes: `- DB_PORT=${DB_PORT:-5432}`
Second, the communication between the PayID and Postgres containers is made via the default bridge network created when we run `npm run devEnvUp` and this communication uses the internal port 5432.
Inside the container, Postgres will always start on port 5432 and we will probably never need to change that.

Here is an example of what happens if we change the DB_PORT inside the payid container:

image

So, to summarize, we need to remove the following two lines in the `payid` service and we should be good to go.
```
      # Local DB_PORT environment variable is used if set (default 5432)
      - DB_PORT='${DB_PORT:-5432}'
```

Please let me know if something is unclear.

docker-compose.yml
```diff
@@ -16,6 +16,8 @@ services:
     depends_on:
       - 'db'
     environment:
+      # Local DB_PORT environment variable is used if set (default 5432)
+      - DB_PORT='${DB_PORT:-5432}'
```
FlorentSG, line 20: This line and the comment above can be removed.

mouradski: Hi @FlorentSG and thank you for the review :) 

yes you're right 👍

mouradski:
@hbergren Thank you, it is I who apologize && thank you for your patience :)

PR #544, opened 2020-06-22
Add `Cache-Control: no-store` to public API

hbergren:

We implicitly don't cache anything because we don't set any cache headers, but we should be _explicit_ that we don't cache things.

We want to not cache things because caching is hard, and we like to be lazy, and payment information should _always_ be up to date.

Also, payments using PayIDs (right now) are not being requested many times in succession, so the benefits of caching are limited.

I learned about the proper value for this header from https://github.com/payid-org/rfcs/issues/23.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

We are now explicit that you should never cache anything the PayID server gives you.

## Test Plan

I wrote a test that verifies the correct value comes back.

[Further Reading: MDN Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control).


src/middlewares/checkPublicApiVersionHeaders.ts
```diff
@@ -74,5 +74,14 @@ export default function checkPublicApiVersionHeaders(
   // Eventually, we'll need to be able to upgrade/downgrade responses.
   res.header('PayID-Version', payIdVersionHeader)
 
+  // TODO:(hbergren) This probably should not live here.
```
austin-king, line 77: good comment

PR #546, opened 2020-06-22
fixing a broken docs link in readme.md

koresar:

fixing a broken docs link in readme.md

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -16,7 +16,7 @@ To clean up the associated Docker containers, you can run `npm run devDown`.
 
 ## Further Reading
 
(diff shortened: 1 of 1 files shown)


hbergren:
Thanks for the fix!

dpurdy:
I had a PR with a fix for this over the weekend and it was blocked on a code owner. can you tell me what we missed?

dpurdy:
See what happened here. There were two different bugs in the same doc. :-) Do we have a simple link checker for docs like this? I feel stupid for not grepping or running the whole doc through a linter?

PR #563, opened 2020-06-25
Use Classic Addresses for seed values

hbergren:

We decided the other day that XRPL addresses for a PayID should always return `classic, tag`, so this updates our seed values and tests to reflect that.

### Context of Change

People get really confused when you deal with multiple address formats, especially since X-Addresses aren't something that `rippled` servers actually understand.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

`npm run test` works.


hbergren:
> This will affect Keefer right?

I guess so? (But it shouldn't. They should spin up their own PayID server to test against, and not use our dev environment 😄 )

dino-rodriguez:
Sounds good, I agree. The dev tools are now good enough to do that.

Mainly just double checking to make sure that he is aware (to know what breaks their integration tests)

PR #565, opened 2020-06-26
Metrics refactor

hbergren:

Prepare our metrics code for making things opt-out. Also, add some various tests.

- Add a new ENV var for opting out of pushing metrics
- Raise errors on invalid push metrics configuration instead of just logging a WARN
- Refactor things a bit and add tests for checking our push metrics configuration

### Context of Change

We want to make pushing metrics opt-out. This is what VS Code and most popular software does, as most people will just stick with the defaults.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Now, we require setting an explicit ENV var if you want to opt-out of pushing metrics.

## Test Plan

Tests added for our parsing of configuration to determine whether or not to push metrics.


src/config.ts
```diff
@@ -29,11 +29,18 @@ const config = {
   },
   metrics: {
     /**
-     * Name of the individual or organization that operates this PayID server.
+     * Whether or not to report PayID server metrics. Defaults to true.
+     * To opt out,  you can set the PUSH_PAYID_METRICS to any value that isn't "true".
+     */
+    reportMetrics: process.env.PUSH_PAYID_METRICS
+      ? process.env.PUSH_PAYID_METRICS === 'true'
+      : true,
```
hbergren, line 37: This is how you will "opt-out" of pushing metrics.

src/config.ts
```diff
@@ -29,11 +29,18 @@ const config = {
...
+    /**
+     * Domain name that operates this PayID server.
      * Used to identify who is pushing the metrics in cases where multiple PayID servers are pushing to a shared metrics server.
      * Required for pushing metrics.
      */
-    organization: process.env.PAYID_ORG,
+    domain: process.env.PAYID_DOMAIN,
```
hbergren, line 43: We want to standardize this to the domain name where the PayID server is running, so rename things to be more explicit about that.

Note that this is _technically_ a breaking change to partners, but I have a trick up my sleeve in the following PR that will be stacked on this, making this not (at least less of) a breaking change.

src/services/metrics.ts
```diff
@@ -30,61 +30,30 @@ const payIdGauge = new Gauge({
   registers: [payIdGaugeRegistry],
 })
 
-/**
- * Determines whether push metrics are enabled by looking at config / environment variables.
- * Used to determine whether we should schedule pushing events for metrics.
- *
- * @returns A boolean for whether push metrics are enabled.
- */
-function isPushMetricsEnabled(): boolean {
```
hbergren, line 39: I just moved this around (I like helpers at the bottom of the file), and I did refactor some of the implementation.

Mostly, wherever it is logging an warning, I now throw an error. That way, if something is misconfigured and you haven't opted out of pushing metrics, you can't even run the PayID server.

src/services/metrics.ts
```diff
@@ -30,61 +30,30 @@ const payIdGauge = new Gauge({
...
-
-  return true
-}
-
 /**
  * Attempt to schedule a recurring metrics push to the metrics gateway URL.
  * Configured through the environment/defaults set in the PayID
  * app config.
  *
+ * @param metricsConfig - A configuration object controlling how to push and generate metrics.
```
hbergren, line 38: Expose this as a parameter for testing purposes.

src/services/metrics.ts
```diff
@@ -30,61 +30,30 @@ const payIdGauge = new Gauge({
...
+ *
  * @returns A timer object if push metrics are enabled, undefined otherwise.
  */
-export function scheduleRecurringMetricsPush(): NodeJS.Timeout | undefined {
-  if (!isPushMetricsEnabled()) {
+export function scheduleRecurringMetricsPush(
+  metricsConfig = config.metrics,
+): NodeJS.Timeout | undefined {
+  if (!arePushMetricsEnabled(metricsConfig)) {
```
hbergren, line 45: Also pass it as a parameter here for testing purposes.

src/services/metrics.ts
```diff
@@ -122,7 +91,67 @@ export function scheduleRecurringMetricsPush(): NodeJS.Timeout | undefined {
...
+ * @param metricsConfig - A configuration object controlling how metrics are pushed and generated.
+ *
+ * @returns A boolean for whether push metrics are enabled.
+ *
+ * @throws An error if pushing metrics is enabled, but the required configuration to push metrics is missing or malformed.
+ */
+function arePushMetricsEnabled(metricsConfig: typeof config.metrics): boolean {
+  if (!metricsConfig.reportMetrics) {
+    return false
+  }
```
hbergren, line 112: This lets you opt out of pushing metrics.

src/services/metrics.ts
```diff
@@ -122,7 +91,67 @@ export function scheduleRecurringMetricsPush(): NodeJS.Timeout | undefined {
...
+    new URL(metricsConfig.gatewayUrl)
+  } catch {
+    throw new Error(
+      `Push metrics are enabled, but the environment variable PUSH_GATEWAY_URL is not a valid url: "${metricsConfig.gatewayUrl}".`,
+    )
+  }
+
+  // If the organization is undefined, just return false.
+  // Because we can wait for the first request that hits the Public API,
+  // and grab the organization off of that.
```
hbergren, line 125: This is not implemented in _this_ PR, but this is how I plan on making the `PAYID_ORG` to `PAYID_DOMAIN` environment variable name change non-breaking.

Basically, we have an environment variable as well... but we can also just use public API requests that hit the PayID server to figure out what domain we are running on.

We could potentially even remove the `PAYID_DOMAIN` environment variable, but I'm open to leaving it as a safety valve.

src/services/metrics.ts
```diff
@@ -122,7 +91,67 @@ export function scheduleRecurringMetricsPush(): NodeJS.Timeout | undefined {
...
+  if (metricsConfig.domain === '') {
+    throw new Error(
+      `
+      Push metrics are enabled, but the environment variable PAYID_DOMAIN is an invalid value (empty string).
+      Please set PAYID_DOMAIN to the domain your PayID server is running on, like "example.com",
+      or you can disable pushing metrics by setting PUSH_PAYID_METRICS to false.
+
+      Metrics only capture the total number of PayIDs by (paymentNetwork, environment),
+      and the (paymentNetwork, environment) of requests to the PayID server.
+      No identifying information is captured.
```
hbergren, line 139: I thought it was good to be explicit about how to disable / enable metrics here and what we were collecting, but let me know your thoughts.

This should basically never happen, as it requires EXPLICITLY setting the PAYID_DOMAIN environment variable to an empty string.

src/services/metrics.ts
```diff
@@ -122,7 +91,67 @@ export function scheduleRecurringMetricsPush(): NodeJS.Timeout | undefined {
...
+      `,
+    )
+  }
+
+  const oneDayInSeconds = 86400
+  if (
+    metricsConfig.pushIntervalInSeconds <= 0 ||
+    metricsConfig.pushIntervalInSeconds > oneDayInSeconds
```
hbergren, line 147: I added this as a reasonable upper limit. If you _are_ pushing metrics, you should probably at least push them daily. (I think the default is 15 seconds).

src/services/payIdReport.ts
```diff
@@ -20,19 +20,26 @@ export async function generatePayIdCountMetrics(): Promise<void> {
...
  * @returns A timer that will generate PayID count metrics every X seconds.
+ *
+ * @throws An error if PAYID_COUNT_REFRESH_INTERVAL is invalid (negative number, or interval greater than 24 hours).
  */
-export default function scheduleRecurringPayIdCountMetrics():
-  | NodeJS.Timeout
-  | undefined {
+export default function scheduleRecurringPayIdCountMetrics(
+  metricsConfig = config.metrics,
+): NodeJS.Timeout | undefined {
```
hbergren, line 31: Refactored this to take `config.metrics` mostly for consistency.

test/unit/pushMetricsConfiguration.test.ts
```diff
@@ -0,0 +1,160 @@
+import 'mocha'
+import { assert } from 'chai'
+
+import config from '../../src/config'
+import { scheduleRecurringMetricsPush } from '../../src/services/metrics'
+import { structuredClone } from '../helpers/helpers'
+
+let recurringPushMetricsTimeout: NodeJS.Timeout | undefined
+
+describe('Push Metrics Configuration - scheduleRecurringMetricsPush()', function (): void {
```
hbergren, line 10: Here we are testing `scheduleRecurringMetricsPush` throws errors appropriately when configuration is invalid. We _could_ test the actual `arePushMetricsEnabled` function, but that isn't exported right now, and it's a little weird to export things just for testing.

Ideally, we just test the visible behavior of publicly accessible functions, and here the visible behavior is the return values and errors & error messages thrown by this function.

src/config.ts
dino-rodriguez: Do you think it's worth changing this variable name? Even if it's not breaking because of the clever idea described below, the partners that want to be explicit and set this would have to go through and re-set it.

src/services/metrics.ts
```diff
@@ -30,61 +30,30 @@ const payIdGauge = new Gauge({
...
  * app config.
  *
+ * @param metricsConfig - A configuration object controlling how to push and generate metrics.
+ *
  * @returns A timer object if push metrics are enabled, undefined otherwise.
  */
-export function scheduleRecurringMetricsPush(): NodeJS.Timeout | undefined {
-  if (!isPushMetricsEnabled()) {
+export function scheduleRecurringMetricsPush(
+  metricsConfig = config.metrics,
```
dino-rodriguez, line 43: Nice, happy we are keeping the global access at the top layer here.

src/services/metrics.ts
```diff
@@ -122,7 +91,67 @@ export function scheduleRecurringMetricsPush(): NodeJS.Timeout | undefined {
...
+ * @throws An error if pushing metrics is enabled, but the required configuration to push metrics is missing or malformed.
+ */
+function arePushMetricsEnabled(metricsConfig: typeof config.metrics): boolean {
+  if (!metricsConfig.reportMetrics) {
+    return false
+  }
+
+  try {
+    // eslint-disable-next-line no-new -- We are using Node's URL library to see if the URL is valid.
+    new URL(metricsConfig.gatewayUrl)
```
dino-rodriguez, line 116: Nice, this looks like a really straightforward way to check for a URL.

Is there a way to avoid the warning it would otherwise raise?

src/services/metrics.ts
```diff
@@ -122,7 +91,67 @@ export function scheduleRecurringMetricsPush(): NodeJS.Timeout | undefined {
...
+  } catch {
+    throw new Error(
+      `Push metrics are enabled, but the environment variable PUSH_GATEWAY_URL is not a valid url: "${metricsConfig.gatewayUrl}".`,
+    )
+  }
+
+  // If the organization is undefined, just return false.
```
dino-rodriguez, line 123: This is clever! I like it.

src/services/metrics.ts
dino-rodriguez: I love this approach. It's really clever.

If we wanted to leave this as a safety valve, I think it would be worth:

- Checking if it is a valid URL (like you do for the `metricsGatewayUrl`)
- If it is, we go with the value from the variable. If it is not a valid URL, or is not set, we go for the value retrieved from PayID requests.

This would theoretically allow us to auto-switch all partners to URLs, while also allowing them to specify it manually if they have a particular org URL they would like to use.

Good idea on this, Hans!

src/services/metrics.ts
```diff
@@ -122,7 +91,67 @@ export function scheduleRecurringMetricsPush(): NodeJS.Timeout | undefined {
...
+  // and grab the organization off of that.
+  if (metricsConfig.domain === undefined) {
+    return false
+  }
+
+  if (metricsConfig.domain === '') {
+    throw new Error(
+      `
+      Push metrics are enabled, but the environment variable PAYID_DOMAIN is an invalid value (empty string).
+      Please set PAYID_DOMAIN to the domain your PayID server is running on, like "example.com",
```
dino-rodriguez, line 134:
```suggestion
      Please set PAYID_DOMAIN to the domain your PayID server is running on, like "https://example.com",
```

test/unit/pushMetricsConfiguration.test.ts
dino-rodriguez: I know we test export some helpers & test directly in other `utils` throughout the code.

I think even if it's a little weird, I would rather test the helpers than not testing them at all. If the helper is wrong, the larger function will be wrong. And you can either solve that by implicitly or explicitly testing the helper.

In saying implicitly, I mean passing in inputs that will invoke the branches in the helper function from public function. The last time we talked, we didn't like this as much as being explicit, though.

src/config.ts
hbergren: I'm actually going to remove this entirely.

We shot ourselves in the foot by setting `PAYID_ORG=payid-demo` as some default value for something and we were spamming ourselves with useless metrics.

So, I'm just going to make this use the `req.hostname` from requests from the public API.

test/unit/pushMetricsConfiguration.test.ts
hbergren: Yeah, I change this in the next PR, but I refactor a _bunch_ of things, so I'm going to leave this as is for this part of the stack.

src/services/metrics.ts
hbergren: You mean the ESLint warning?

I can't think of anything. You can't call `URL` without the `new` keyword, and if we pass it to a variable, that variable would be unused and ESLint gets angry.

src/services/metrics.ts
hbergren: Nope. Here's the example from the Express documentation:
```ts
console.dir(req.hostname)
// => 'example.com'
```

Also, we don't actually care about the `https://` when we collect metrics.

src/services/payIdReport.ts
```diff
@@ -20,19 +20,26 @@ export async function generatePayIdCountMetrics(): Promise<void> {
...
   const refreshIntervalInSeconds =
-    config.metrics.payIdCountRefreshIntervalInSeconds
+    metricsConfig.payIdCountRefreshIntervalInSeconds
 
-  if (refreshIntervalInSeconds <= 0 || Number.isNaN(refreshIntervalInSeconds)) {
-    logger.warn(
-      `Invalid PAYID_COUNT_REFRESH_INTERVAL value: ${refreshIntervalInSeconds}. PayID count metrics will not be scheduled.`,
+  const oneDayInSeconds = 86400
```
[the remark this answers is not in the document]
hbergren: Yeah, so in my next PR I actually merge `payIdReport` into the `metrics` file, so the const lives in one place.

> As an aside -- should we use PayID style errors for this? I noticed we are just using Error.

Not sure what error to use. It's not a `ParseError`. Open to ideas?

I think it's fine though. We run this at the very beginning and it's not really a "PayIdError" per se, it's more of a "app initialization check".

src/services/metrics.ts
austin-king: Ya, think opting to be a bit verbose on informing people of how / where we're collecting metrics is the right choice

src/services/payIdReport.ts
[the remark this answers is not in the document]
austin-king: I think it'd be fine to add an error for app initialization errors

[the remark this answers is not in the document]
dino-rodriguez: Yeah, it's not an error that happens during the usage of the protocol

I think a an 'initialization' error or 'startup' error wouldn't hurt, if you wanted to go for that

PR #567, opened 2020-06-26
chore+feat: script to compare versions

dino-rodriguez:

Adds a shell script to compare the Docker, NPM and Git tag versions.

If there is a mismatch, this script errors and exits.

If they all match, it prints out the current version.

Adds this shell script as an NPM script.

### Context of Change

Currently all of our versioning is manual.

We need to keep 3 versions, and these should stay synced.

Currently, they are not in sync.

To start automating this process, I wrote a script that can compare these
versions.

I looked for something comparable for our use case, and could not find it, so
decided to roll my own.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Adds a shell script & npm script

## Test Plan

Tested locally


version.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh
+
+# Parses a YAML file and prints the key value pairs.
+# Adapted from: https://gist.github.com/pkuczynski/8665367
+#
+# $1 - The YAML file path string.
+# $2 - The prefix string for the outputed keys.
```
hbergren, line 7:
```suggestion
# $2 - The prefix string for the outputted keys.
```

version.sh
```diff
@@ -0,0 +1,57 @@
...
+   awk -F$fs '{
+      indent = length($1)/2;
+      vname[indent] = $2;
+      for (i in vname) {if (i > indent) {delete vname[i]}}
+      if (length($3) > 0) {
+         vn=""; for (i=0; i<indent; i++) {vn=(vn)(vname[i])("_")}
+         printf("%s%s%s=\"%s\"\n", "'$prefix'",vn, $2, $3);
+      }
+   }'
+}
```
hbergren, line 24: What does this black magic do?

version.sh
```diff
@@ -0,0 +1,57 @@
...
+      error "Version Mismatch:
+
+      Docker Image Version: "$docker_image_version"
+      NPM Version: "$npm_version"
+      Git Tag Version: "$git_tag_version"
+      "
+   else
+      echo "Docker, NPM, Git Tag Version: "$npm_version
+   fi
+}
```
hbergren, line 55: This I get at a high level (although not the shell specifics). I like it!

dino-rodriguez: This function basically gets the version from each of those (using cut to parse out extraneous information from the version string that is not just x.y.z).

version.sh
dino-rodriguez: So high level:

`sed` is a stream editor (often used to substitute expressions)

`awk` is a scripting language used for data extraction (e.g. handling `csv` files and the like)

Im not sure on the exact specifics of the regex call there, but _basically_ what this is doing is:

- `sed` is swapping out certain characters for field separators
- `awk` receives the piped stream output, and because of the field separators, can iterate through it and generate key value pairs, which get printed to the console
- the key value pairs looks like key=value, so when you run `eval` on the output of this, you load all of the key/value pairs into the shell environment
- the result is that you can reference the keys in the yaml file

package.json
```diff
@@ -29,7 +29,8 @@
     "testWatch": "nyc mocha --watch 'test/**/*.test.ts'",
     "devEnvUp": "docker-compose up -d",
     "devDbUp": "docker-compose up -d db",
-    "devDown": "docker-compose down"
+    "devDown": "docker-compose down",
+    "compareVersions": "./version.sh $npm_package_version"
```
hbergren, line 33: Add `compareVersions` to `lintNoFix` please. 😄

dino-rodriguez: That's in the next PR. This PR just builds the new stuff, the next one adds it and fixes the old versions.

hbergren:
Question: How will this script fare on Windows 10?

.gitlab-ci.yml
```diff
@@ -1,4 +1,4 @@
-###
+#####
```
austin-king, line 1: What's the reason for these extra `#`?

version.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/sh
+
+# Parses a YAML file and prints the key value pairs.
+# Adapted from: https://gist.github.com/pkuczynski/8665367
+#
+# $1 - The YAML file path string.
+# $2 - The prefix string for the outputted keys.
+parse_yaml() {
+   declare -r prefix=$2
```
austin-king, line 9: Can we get comments on what these are doing?

.gitlab-ci.yml
dino-rodriguez: Whoops, updated. Not sure what happened there.

docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 version: '3'
 services:
-  payid-server:
```
dino-rodriguez, line 3: nit : all the fields in the docker compose use underscores

PR #570, opened 2020-06-26
Metrics refactor 2

hbergren:

Actually makes metrics `opt-out`.

Basically we:
- Eliminate the `PAYID_ORG/PAYID_DOMAIN` env var, and just dynamically compute that from the first request that hits our public API.
- Check if our configuration is valid for metrics when we initialize the app.
- Log a message to the console if we _are_ pushing metrics (don't keep that a secret).

We probably also want to document metrics pushing in the README and the docs site a little better, but... it's 5:30 PM on a Friday, so if you guys have ideas on Monday, I'm happy to discuss.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

`npm run test` works.


.vscode/launch.json
```diff
@@ -8,7 +8,7 @@
       "type": "node",
       "request": "launch",
       "name": "Launch PayID Server",
-      "skipFiles": ["<node_internals>/**"],
+      "skipFiles": ["<node_internals>/**", "node_modules/**/*.js"],
```
hbergren, line 11: I kept running into this problem debugging tests where I would be stepping into Mocha code (obnoxious).

This ignores all code in `node_modules`, so you'll only step through PayID code when debugging.

This is what you'll want 99% of the time, but if you _ever_ want to debug a dependency in the context of PayID, you'll have to come edit this.

package.json
```diff
@@ -25,7 +25,7 @@
     "prestart": "npm run build",
     "start": "node build/src/index.js",
     "startWatch": "tsc-watch --onSuccess 'npm run start'",
-    "test": "nyc mocha 'test/**/*.test.ts'",
+    "test": "nyc mocha 'test/**/*.test.ts' --exit",
```
hbergren, line 28: This is a super-hack because I can't Mocha.

Something in the way I refactored things / rewrote the tests means I have something lying around that should be cleaned up but isn't.

This can happen when you have database connections left over or something like that. I have _no idea_ what this could be, so I'm just calling `--exit` to force Mocha to close things down when all the tests finish running.

This is _not_ the right way to do things, but I honestly can't figure out what is still running.

src/app.ts
```diff
@@ -49,8 +46,14 @@ export default class App {
...
-    this.recurringMetricsPushTimeout = scheduleRecurringMetricsPush()
-    this.recurringMetricsTimeout = scheduleRecurringPayIdCountMetrics()
+    checkMetricsConfiguration(initConfig.metrics)
+
+    // Explicitly log that we are pushing metrics if we're pushing metrics.
+    if (initConfig.metrics.pushMetrics) {
+      logger.info(
+        'Pushing metrics is enabled. If you would like to opt out of pushing metrics, set the environment variable PUSH_PAYID_METRICS to "false".',
+      )
+    }
```
hbergren, line 56: Thought we should explicitly log that we're collecting metrics. We aren't trying to hide this, we're just making it opt-out.

We should _probably_ add this to documentation as well. README? The docs website? Happy for ideas.

src/middlewares/initializeMetrics.ts
```diff
@@ -0,0 +1,31 @@
...
+ * An Express middleware that schedules metrics generation.
+ *
+ * It also looks at the request hostname property to dynamically set
+ * the domain to associate the metrics it pushes with (if push metrics are enabled).
+ *
+ * @param req - An Express request object. We get the hostname off of this for pushing metrics.
+ * @param _res - An Express response object (unused).
+ * @param next - An Express next function.
+ */
+export default function initializeMetrics(
```
hbergren, line 16: This is a new middleware that dynamically sets the `metrics.domain` based on the first public API request and starts pushing metrics if configured to push metrics.

src/services/metrics.ts
```diff
@@ -3,112 +3,262 @@ import { hostname } from 'os'
 import { Counter, Gauge, Pushgateway, Registry } from 'prom-client'
 
 import config from '../config'
+import getPayIdCounts from '../data-access/reports'
 import logger from '../utils/logger'
 
-// Custom Prometheus registries.
```
hbergren, line 8: I totally rewrote this entire file:

The basic structure is now:
```ts
class Metrics {}
const metrics = new Metrics()

export metrics
```

I needed to do this because we now talk to metrics in multiple places (`app.ts`, `initializeMetrics.ts` middleware), and metrics involves a lot of shared state.

So, the easiest thing to do was to bundle up most of that state into a class.

Note that we still hit `config` as a global object here, but I think that's ok.

src/services/payIdReport.ts
```diff
@@ -1,55 +0,0 @@
-import config from '../config'
```
hbergren, line 1: This entire file got deleted and merged into `/services/metrics.ts`.

test/integration/e2e/private-api/metrics.test.ts
```diff
@@ -32,7 +32,7 @@ describe('E2E - privateAPIRouter - GET /metrics', function (): void {
     await createPayId(payId, network, mainnet)
     await lookupPayId(account, asAccept(network, mainnet), HttpStatus.OK)
     await assertMetrics(
-      /payid_lookup_request\{paymentNetwork="ACH",environment="MAINNET",org="undefined",result="found"\} 1/u,
+      /payid_lookup_request\{paymentNetwork="ACH",environment="MAINNET",org="127.0.0.1",result="found"\} 1/u,
```
hbergren, line 35: This is neat. Notice how we had to change the `org` from undefined to `127.0.0.1` because _now_ we dynamically set the metrics domain based on the first public API request.

test/unit/checkMetricsConfiguration.test.ts
```diff
@@ -2,44 +2,88 @@ import 'mocha'
 import { assert } from 'chai'
 
 import config from '../../src/config'
-import { scheduleRecurringMetricsPush } from '../../src/services/metrics'
+import { checkMetricsConfiguration } from '../../src/services/metrics'
 import { structuredClone } from '../helpers/helpers'
 
-let recurringPushMetricsTimeout: NodeJS.Timeout | undefined
+describe('Push Metrics Configuration - checkMetricsConfiguration()', function (): void {
```
hbergren, line 8: Changed this test stuff to actually test the function where this logic lives.

src/app.ts
hbergren: Should we log if metrics are _disabled_?

hbergren: Also, is this the right logging level? I don't think WARN is valid, but maybe I'm wrong?

dino-rodriguez: I think the docs website makes sense.

This log is pretty explicit.

This looks like the right log level to me.

src/app.ts
```diff
@@ -60,13 +63,7 @@ export default class App {
     this.publicAPIServer?.close()
     this.privateAPIServer?.close()
 
-    if (this.recurringMetricsTimeout?.hasRef()) {
```
dino-rodriguez, line 63: Did these get refactored to somewhere?

src/middlewares/initializeMetrics.ts
```diff
@@ -0,0 +1,31 @@
...
+  req: Request,
+  _res: Response,
+  next: NextFunction,
+): void {
+  // Start metrics on the first public API request.
+  // This will _always_ happen at initialization unless the PAYID_DOMAIN env var is set.
+  if (!config.metrics.domain || !metrics.areMetricsRunning()) {
```
dino-rodriguez, line 23: Love this middleware idea!

I think even if this is set, we should check that it is a valid URL, and if it is not a valid URL, we use this instead as well (to avoid receiving non-url org strings).

test/integration/e2e/private-api/metrics.test.ts
dino-rodriguez: That is neat!

src/app.ts
hbergren: Yup, this is in `metrics.stopMetricsGeneration()`. It gets called by `app.close()`.

src/middlewares/initializeMetrics.ts
hbergren: I actually ripped out the ENV var that could set this explicitly, so it's _always_ going to be whatever the first public API request is.

src/app.ts
austin-king: Ya I think we should log on both enabled and disabled

src/app.ts
```diff
@@ -48,9 +45,15 @@ export default class App {
     this.publicAPIServer = this.launchPublicAPI(initConfig.app)
     this.privateAPIServer = this.launchPrivateAPI(initConfig.app)
 
-    // Attempt to schedule recurring metrics.
-    this.recurringMetricsPushTimeout = scheduleRecurringMetricsPush()
-    this.recurringMetricsTimeout = scheduleRecurringPayIdCountMetrics()
+    // Check if our metrics configuration is valid.
+    checkMetricsConfiguration(initConfig.metrics)
```
austin-king, line 49: If this is just validating that metrics are properly configured can we change the name from `checkMetricsConfiguration` to `validateMetricsConfiguration` or `isValidMetricsConfiguration`?

hbergren: This comes down to naming semantics.

I tend to think of `validateX` or `isValidX` as a function that should return `true/false`. But this function returns nothing if it's valid, and throws an error if it's invalid.

Maybe `validateMetrisConfiguration` is the right name for that?

src/middlewares/initializeMetrics.ts
dino-rodriguez: Great !

PR #571, opened 2020-06-26
refactor: bash script into lib

dino-rodriguez:

Turns the `version.sh` into a library. 

Moves to the `scripts` folder.

Update npm script to reference the `compare_versions` function directly
from this library.

### Context of Change

I am going to re-use some of the functions in this file to implement bumping. 

Thus, it makes more sense for this to be a library instead of a script.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Updated bash script to be a library, moved to `scripts`

## Test Plan

Tested locally


scripts/version.sh
```diff
@@ -1,11 +1,36 @@
 #!/bin/sh
 
+# Compares the Docker image version, NPM version, and Git tag version
+# in the repo. Throws an error if any do not match.
+#
+# $1 - The NPM version string.
+function compare_versions() {
```
dino-rodriguez, line 7: This is just moving around functions, and adding the `function` keyword.

scripts/version.sh
```diff
@@ -32,36 +57,7 @@ parse_yaml() {
...
-      NPM Version: "$npm_version"
-      Git Tag Version: "$git_tag_version"
-      "
-   else
-      echo "Docker, NPM, Git Tag Version: "$npm_version
-   fi
-}
-
-# Run the script
-compare_versions $1
```
dino-rodriguez, line 67: This command is the only thing dropped from this file.

package.json
```diff
@@ -30,7 +30,7 @@
     "devEnvUp": "docker-compose up -d",
     "devDbUp": "docker-compose up -d db",
     "devDown": "docker-compose down",
-    "compareVersions": "./version.sh $npm_package_version"
+    "compareVersions": ". ./scripts/version.sh && compare_versions $npm_package_version"
```
hbergren, line 33: Is this `. ./scripts` correct? What is that doing?

scripts/version.sh
```diff
@@ -1,11 +1,36 @@
 #!/bin/sh
 
+# Compares the Docker image version, NPM version, and Git tag version
+# in the repo. Throws an error if any do not match.
+#
+# $1 - The NPM version string.
+function compare_versions() {
+   eval $(parse_yaml docker-compose.yml "config_")
+   declare -r docker_image_version=$(echo $config_services_payid_image | cut -f 2 -d ':')
+   declare -r npm_version=$1
```
hbergren, line 10: Nit: Put this above `docker_image_version`?

scripts/version.sh
```diff
@@ -1,11 +1,36 @@
...
+   declare -r git_tag_version=$(git describe --tags | cut -f 1 -d '-')
+
+   if [[ $docker_image_version != $npm_version || \
+         $docker_image_version != $git_tag_version || \
+         $npm_version != $git_tag_version
+   ]]; then
```
hbergren, line 16: I guess I'm ok with this, because it's explicit we're asserting they are all equal to each other, but since equality is transitive, we could just say:
`docker != npm || git != npm`, and that would cover all our bases.

package.json
austin-king: also can we run `compare_versions` right after the `version.sh` script or do we also need to change the path down into the scripts directory to execute `compare_versions`?

dino-rodriguez: The `version.sh` file just has bash functions in it and it doesn't actually execute any of the functions in the file

So we need to source those functions with `. ./scripts/version.sh`

And then call `&& compare_versions` to run that function and exit if it fails

All of these PRS I tested locally, but it would be good if one of you guys could test locally on your machine to confirm.

scripts/version.sh
dino-rodriguez: Ok, I went ahead and updated it

docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
     # Defines the container name when using "docker-compose up". Warning: does not work with "docker-compose run" (Docker issue)
     container_name: 'payid-server'
     # Defines the image name and tag.
-    image: payid-server:1.0.1
+    image: payid:${PAYID_SERVER_IMAGE_TAG}
```
hbergren, line 7: Can we rename this to `PAYID_DOCKER_IMAGE_VERSION`?

package.json
```diff
@@ -27,10 +27,11 @@
...
     "test": "nyc mocha 'test/**/*.test.ts'",
     "testWatch": "nyc mocha --watch 'test/**/*.test.ts'",
-    "devEnvUp": "docker-compose up -d",
-    "devDbUp": "docker-compose up -d db",
-    "devDown": "docker-compose down",
-    "compareVersions": "./version.sh $npm_package_version"
+    "devEnvUp": "PAYID_SERVER_IMAGE_TAG=$npm_package_version docker-compose up -d",
+    "devDbUp": "PAYID_SERVER_IMAGE_TAG=$npm_package_version docker-compose up -d db",
+    "devDown": "PAYID_SERVER_IMAGE_TAG=$npm_package_version docker-compose down",
+    "compareVersions": "bash -c '. ./scripts/version.sh && compare_versions $npm_package_version'",
```
hbergren, line 33: Do you need to do `npm run compareVersions` above? I would think you could just do `compareVersions`? (In `lint` and `lintNoFix`).

package.json
```diff
@@ -27,10 +27,11 @@
...
+    "bump": "bash -c '. ./scripts/version.sh && bump $npm_package_version'"
```
hbergren, line 34: So how does this work? Do I need to change `package.json`s version and then I run `npm run bump`?

dino-rodriguez: You just need to run `npm run bump`

`$npm_package_version` is the current version

It used inside the bump script to run `compareVersions` first, validate that all versions are equal before bumping anything

package.json
dino-rodriguez, line 33: We have to explicitly invoke `bash` here. CI was failing because of this.

By default, CI uses `sh`. We need to explicitly do it here ( instead of at the top of the `version.sh` script ), because we are using `version.sh` as a library, as opposed to executing `compare_versions`.

hbergren: Would it be easier if we didn't have to source the functions in `version.sh` and we just split that out into a few separate shell scripts?

dino-rodriguez: We need to do `npm run compareVersions` ( just tried this locally with just `compareVersions`)

This happens because we are running the npm script

If we wanted to run the bash script directly, we would have to source it as well

dino-rodriguez: I think that could be another way to go. For now, I don't mind having one file and sourcing the library ( since we also use `compareVersions` inside of bump, so we would have to source that as well.

PR #572, opened 2020-06-27
chore: version bump script

dino-rodriguez:

Adds a new npm script `npm run bump` which accepts one parameter (major, minor,
or patch) to bump the version in NPM, cut a new git tag, & updated the version
in the `docker-compose.yml`.

### Context of Change

Allows you to bump the npm version alongside the docker version to keep them in
sync.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Adds bump script 

## Test Plan

Tested locally -- cut a patch present in this PR


docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
     # Defines the container name when using "docker-compose up". Warning: does not work with "docker-compose run" (Docker issue)
     container_name: 'payid'
     # Defines the image name and tag.
-    image: payid:1.0.1
+    image: payid:1.0.2
```
dino-rodriguez, line 7: It works ~

scripts/version.sh
```diff
@@ -1,5 +1,29 @@
...
+         $bump_increment != "patch"
+   ]]; then
+      error "Invalid bump increment. Please specify 'major', 'minor', or 'patch'."
+   else
+      # Bump NPM version and git tag
+      declare -r new_version=$(npm version $bump_increment)
+
+      # Bump the Docker image version by finding and replacing the old
+      # version for the new. Drop the 'v' from the Git tag.
+      sed -i "" "s|payid:$current_version|payid:${new_version:1}|g" docker-compose.yml
```
dino-rodriguez, line 23: The git tags autogenerate as `v1.0.0`

The npm tags generate without the `v`

We make sure to cut out the `v` when doing any comparisons or lookups

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "payid",
-  "version": "1.1.0",
+  "version": "1.0.2",
```
hbergren, line 3: Why on earth was this ever `1.1.0`?

hbergren: Also, how do you bump the `package-lock.json`? Do you run `npm install` after bumping `package.json`?

hbergren:
Awesome. This definitely gets us 80% of the way there!

package-lock.json
dino-rodriguez: Hahahaha

That was because I was doing a bunch of testing locally, and was bumping versions

I'll edit the other branch to `1.0.1` on the `package-lock.json` (the branch this is merging into) as that seems to be what is causing the issue

Nice catch! Damn. Did you manually open the `package-lock.json`?

And the `package-lock.json` gets bumped automatically when you run `npm version` :)

FlorentSG:
Hey @dino-rodriguez , @hbergren , I had a quick look at this and I wanted to share my idea regarding the `payid-server` image tag in `docker-compose.yml`.

In `docker-compose.yml` you can write the `tag` as a variable, so that would be:
`image: payid-server:${PAYID_SERVER_IMAGE_TAG}`

and in `package.json` you can pass that variable to the docker-compose command like that:
`"devEnvUp": "PAYID_SERVER_IMAGE_TAG=$npm_package_version docker-compose up -d",`

I just tested it and the payid-server image is correctly tagged after updating the version in package.json.

Glad to share if that can help :)

hbergren:
> Hey @dino-rodriguez , @hbergren , I had a quick look at this and I wanted to share my idea regarding the payid-server image tag in docker-compose.yml.

That's awesome! @dino-rodriguez , I would definitely recommend using that approach if we can. That way we don't need the YAML parsing in the shell script (less code, less complex). What do you think?

docker-compose.yml
austin-king: I would think the addition of an npm script update shouldn't be an increment to the actual versioning. Can we decrement this version back to 1.0.1?

dino-rodriguez: I bumped this to 1.0.2 to both :

- Test this script to confirm it's working
- I figured we needed to cut a release

I agree that should probably live in it's own PR, so I'm removing this

dino-rodriguez:
@FlorentSG @hbergren 

Florent thank you for the suggestion! That really simplifies things. I went with this.

PR #573, opened 2020-06-27
chore: add release script

dino-rodriguez:

Adds a `npm run release` script to create a draft release.

Also adds `bump` script.

### Context of Change

Rounds out a basic set of release tooling for PayID.

Cutting a release should now be as easy as:

`npm run bump x` (x = major, minor, patch)
`npm run release`

And then you can go to the release draft and update the contents before
publishing.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Adds release script 

## Test Plan

Tested locally


package.json
```diff
@@ -31,7 +31,8 @@
     "devDbUp": "docker-compose up -d db",
     "devDown": "docker-compose down",
     "compareVersions": ". ./scripts/version.sh && compare_versions $npm_package_version",
-    "bump": ". ./scripts/version.sh && bump $npm_package_version"
+    "bump": ". ./scripts/version.sh && bump $npm_package_version",
+    "release": "hub release create -d -m $npm_package_version $npm_package_version"
```
dino-rodriguez, line 35: Requires that the `hub` dependency be installed locally

This can be installed via `brew install hub`

If this fails the first time you install, it's probably because you need to run the first time setup to link it to the repo/your github account

hbergren: I would prefer we didn't depend on globally installed dependencies like `hub`.

Do you think we could use a GitHub Action for this like I suggested in this PR:
https://github.com/xpring-eng/xpring-common-js/pull/265/files?

I think we can write a GH action such that, when you push up a `git tag`, it automatically creates a draft GH release.

dino-rodriguez: So the only people that will really be depending on this will be either you / me / or austin for cutting releases for the foreseeable future

I would definitely not want to use a global dep. if it would affect other devs, but this script is a bit of a special case

That being said, I can look into the other option 

The release stuff is more github-ee, so that would be a good take off point to start including things in CI as opposed to locally (we haven't had to deal with any github stuff up until this script)

hbergren: Fair. I guess I'm ok with having a global dependency on hub if it's just you, me, and Austin. I hadn't considered that perspective.

PR #576, opened 2020-06-29
Change Docker image name

hbergren:

Change the container name from `payid` to `payid-server`.

### Context of Change

Someone posted an issue asking whether `npm run devEnvUp` created a PayID in their database/server, because the output of Docker is literally `'Creating payid'`. I don't think that's an unreasonable takeaway given that message.

So, we update our Docker image name to `payid-server` to avoid confusion moving forward.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Ran `npm run devEnvUp` and things looked ok.

It's a name change.


docker-compose.yml
```diff
@@ -1,20 +1,20 @@
-version: '3'
+version: "3"
```
hbergren: I think this stuff is just Prettier deciding that double quotes are better.

This should _really_ be another PR, but it's not a big diff, so I'm leaving it.

docker-compose.yml
```diff
@@ -1,10 +1,10 @@
 version: '3'
 services:
-  payid:
+  payid-server:
     # Defines the container name when using "docker-compose up". Warning: does not work with "docker-compose run" (Docker issue)
-    container_name: 'payid'
+    container_name: 'payid-server'
     # Defines the image name and tag.
-    image: payid:1.0.0
+    image: payid-server:1.0.0
```
hbergren, line 7: This is the meat of the PR.

Rename the service, container name, and image name to `payid-server` to avoid confusion.

hbergren:
Tagging @FlorentSG for review as our resident Docker expert.

Issue #539, opened 2020-06-20
Win 10 Pro Docker install - payid missing file error 127


solonmedia:
## Expected Behavior

Install Docker for Windows
From PowerShell adm run devEnvUp
Install and run both payid and payid-database
Both machines run

## Actual Behavior

At step 4, payid does not start and yields the following erro:

/usr/local/bin/docker-entrypoint.sh: exec: line 8: /opt/payid/scripts/wait-for-postgres.sh: not found

## Environment

- Node version: 12.18.1
- NPM version: 6.14.5
- Operating System and version: Windows 10 Pro v 2004
- PayID server version: 1.0.0
- PayID Version header (if applicable):

dino-rodriguez:
what directory are you running these steps from?

hbergren:
Hey @solonmedia, someone on our team did a fresh Windows 10 pro v 2004 Bootcamp partition, installed Docker and Node.js 12, and PayID ran successfully after `npm run devEnvUp`.

Can you give us more guidance to reproduce the issue?

solonmedia:
@dino-rodriguez I've cloned the git in payid in a windows user subfolder

N:\Users\suppo\Dockers\payid\

I am running the npm run commands inside that payid folder.

npm run devEnvUp

Starting payIid-database ... done
Recreating payid ... done

And in the docker setup for payid, postgres is running fine - it's the payid container that will not start.

Port 8080 is clear. Just seems strange that the composer ran without errors and postgres is running - it's just the payid that errors out with the missing file complaint.

I can try cloning the repo on a non-system drive, but I'm in an admin account.

In the log screen for the payid container, the only payid messages are:

payid | /usr/local/bin/docker-entrypoint.sh: exec: line 8: /opt/payid/scripts/wait-for-postgres.sh: not found
payid | /usr/local/bin/docker-entrypoint.sh: exec: line 8: /opt/payid/scripts/wait-for-postgres.sh: not found

FlorentSG:
Hi @solonmedia , thank you for your response.

After cloning the PayID repository, are you modifying any file or executing any command other than `npm run devEnvUp`?
Before running `npm run devEnvUp` can you confirm the `scripts` directory is there with `wait-for-postgres.sh` in it?
Can you show us the content of `.dockerignore` in the root payid directory, please?
When the `payid` container is created, does it keep running or does it stop?

If you can include some screenshots that would be helpful, thank you.

solonmedia:
[...]

PR #577, opened 2020-06-29
Windows 10 line endings

hbergren:

Add a `.gitattributes` file to [avoid Windows 10 errors on automatic line ending conversions](https://github.com/payid-org/payid/issues/539#issuecomment-651144327).

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

I don't have Windows 10, so I didn't test 😄

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,16 @@
-# Always convert line endings to LF on checkout, even on Windows.
-* text eol=lf
+# Set default behavior to automatically normalize line endings.
+# This will treat all files as text files,
+# and convert to OS's line ending on checkout and back to LF on commit automatically.
+* text=auto
+
+# Behavior for images and documents, which are treated as binary and shouldn't be normalized.
+*.jpg   binary
+*.jpeg  binary
+*.png   binary
+*.gif   binary
+*.ico   binary
+*.pdf   binary
+
+# Force shell scripts to always use LF line endings.
+# Necessary because we run our shell scripts in Docker UNIX containers.
+*.sh text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
-text eol=lf
+# Always convert line endings to LF on checkout, even on Windows.
+* text eol=lf


hbergren:
Adding @FlorentSG as a reviewer since he was able to replicate the initial Windows 10 problem.

@FlorentSG , is this the right fix at the repository level?

FlorentSG:
@hbergren , from what I can read on different websites, including best practices, I think we should be a bit more specific.
It looks like some pictures are sometimes changed where they shouldn't for example (we don't want our logo to become unreadable).

Here is what I propose for .gitattributes, it is read and applied from top to bottom:

```
# Set default behavior to automatically normalize line endings.
* text=auto

# Behavior for image files, which are treated as binary by default.
*.jpg   -text
*.jpeg   -text
*.png   -text
*.gif   -text

# Behavior for document files, which are treated as binary by default.
*.pdf   -text

# Force bash scripts to always use LF line endings
*.sh text eol=lf
```

Here are the sites I've checked:
https://git-scm.com/docs/gitattributes
https://rehansaeed.com/gitattributes-best-practices/
https://github.com/dotnet/corefx/blob/master/.gitattributes
https://dev.to/deadlybyte/please-add-gitattributes-to-your-git-repository-1jld

Let me know what you think.

hbergren:
@FlorentSG , I have a few questions about your `.gitattributes` proposal:

- Why set `text=auto` instead of `lf` for all files?
- What does that behavior for image files and PDFs do? (It seems like the best solution would be to use Git LFS, which we aren't using).

FlorentSG:
Hi @hbergren , so when the issue happened for both users on Windows, the problem was only with the shell scripts.
At this point, I don't see the need to convert all files to `LF` at checkout on all platforms (except for `.sh`).

`* text=auto` => This will treat all files as text files and convert to OS's line ending on checkout and normalized back to `LF` on commit automatically. On Windows, line-endings are terminated with a combination of a carriage return (ASCII 0x0d or \r) and a newline(\n), also referred to as CR/LF. I think we should leave the default OS behavior as much as possible as long as PayID can start successfully.

From what I read, Git seems "smart" enough to understand which files are text (and need to be normalized) and which ones are binaries (such as images) that don't need to be changed. 
Nevertheless, I read that sometimes Git makes a mistake and tries to normalize a binary file. As far as I know, we only have one binary file at the moment, which is the PayID logo. 
We can clearly specify Git to not touch these files during commit and checkout. It seems that `binary` can be used and is an alias for `-text -diff`

So my new proposition is:

```
# Set default behavior to automatically normalize line endings.
# This will treat all files as text files and convert to OS's line ending on checkout and back to LF on commit automatically.
* text=auto

# Behavior for images and documents, which are treated as binary and shouldn't be normalized.
*.jpg   binary
*.jpeg  binary
*.png   binary
*.gif   binary
*.pdf   binary

# Force shell scripts to always use LF line endings
*.sh text eol=lf
```

https://help.github.com/en/enterprise/2.15/user/articles/configuring-git-to-handle-line-endings

Please let me know what you think. Thanks.

hbergren:
Yeah, that's a great explanation. My main concern was checking CRLF lines into the repository itself, but if Git can automatically convert on checkout and checkin, then I have nothing to worry about.

I think adding a comment to that effect in `.gitattributes` is really helpful though, especially for people who aren't intricately familiar with Git.

I'll update my PR to reflect your new proposal.

Issue #142, opened 2020-03-24
Implement `PATCH` for `/users/{payid}`


hbergren:
Right now, it’s a huge hassle to update someone’s PayID. You have to `GET xpring.money:8081/hbergren$xpring.money`

Which gives you a payload like this:
```
{
  "payId": "hbergren$xpring.money",
  "addresses": [...],
}
```

And then you have to update the `payId` field in that response payload, and send it back to the server using:
`PUT xpring.money:8081/hbergren$xpring.money` (including the payload).

### Proposed Solution

What would be great is to have something like this:

```
PATCH xpring.money:8081/hbergren$xpring.money

{
  "payId": "hansbergren$xpring.money"
}
```

And have that just update the PayID of the user.

#### Requirements

- Updating the `payId` of a user should result in a `201 - Created` response
- Because the resource lives at a new place, we should include a `Location` header
- E2E tests for the `PATCH` method should be implemented
- We should require a request header `Content-Type: application/merge-patch+json`
- The Admin API server should set a `Accept-Patch: application/merge-patch+json` header.

FlorentSG:
I am taking this one, I created the branch `fuzio-patch-user-payid-issue142` for it.

hbergren:
Closed by #586

PR #586, opened 2020-07-07
Admin API - PATCH PayID

FlorentSG:

Addresses the issue 142 and creates a new Admin PATCH API to update a PayID only.

### Context of Change

It is a new feature. I followed the requirements in https://github.com/payid-org/payid/issues/142
I would like to check with you the following points:

- I implemented the required header in `checkAdminApiVersionHeaders`, do you think it is the right place for it?
- Before implementing the PATCH route to update the addresses only, as mentioned in https://github.com/payid-org/payid/issues/142, I think it would be useful to work on the PayID validation (in the URL and in the body). Many endpoints need that validation. We are duplicating code which is not ideal.
I worked on the validation in https://github.com/payid-org/payid/pull/320 but I don't know if it is possible to recover what I've done there. Do you know?

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

A new endpoint is available:
`PATCH /users/:payId`

## Test Plan

I ran `npm run test` and did different test by starting the application with `npm start`


src/data-access/users.ts
```diff
@@ -41,6 +41,31 @@ export async function insertUser(
...
+ *
+ * @param oldPayId - The current PayID which needs to be updated.
+ * @param newPayId - The new PayID of the user.
+ *
+ * @returns The updated user Account.
+ */
+export async function replacePayId(
+  oldPayId: string,
+  newPayId: string,
+): Promise<readonly Account[]> {
```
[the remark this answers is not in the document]
FlorentSG: Good point thanks, there is no need of an Array indeed, I will update that.

src/data-access/users.ts
```diff
@@ -41,6 +41,31 @@ export async function insertUser(
...
+  return knex.transaction(async (transaction: Transaction) => {
+    const updatedAccount = await knex<Account>('account')
+      .where('payId', oldPayId)
+      .update({ payId: newPayId })
+      .transacting(transaction)
```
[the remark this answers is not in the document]
FlorentSG: I am discovering Knex so I am relying on your opinion for this one :)
The new query looks much simpler without the `transaction` stuff.
Simpler is often better so I am happy with the change.

src/middlewares/checkAdminApiVersionHeaders.ts
```diff
@@ -56,3 +56,34 @@ export default function checkAdminApiVersionHeaders(
...
+ * @throws A ParseError if the Content-Type header is missing, malformed, or unsupported.
+ */
+export function checkAdminPatchApiHeaders(
+  req: Request,
+  res: Response,
+  next: NextFunction,
+): void {
+  // Add this header to all successful responses.
+  // We add it early so even errors will respond with Accept-Patch header.
+  const patchRequestHeaderValue = 'application/merge-patch+json'
```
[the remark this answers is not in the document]
FlorentSG: I renamed `patchRequestHeaderValue` to `patchContentType` and added some new comments, please let me know if you agree with the formulation :)

src/middlewares/checkAdminApiVersionHeaders.ts
```diff
@@ -56,3 +56,34 @@ export default function checkAdminApiVersionHeaders(
...
+ *
+ * @throws A ParseError if the Content-Type header is missing, malformed, or unsupported.
+ */
+export function checkAdminPatchApiHeaders(
+  req: Request,
+  res: Response,
+  next: NextFunction,
+): void {
+  // Add this header to all successful responses.
+  // We add it early so even errors will respond with Accept-Patch header.
```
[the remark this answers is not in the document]
FlorentSG: Done

src/middlewares/checkAdminApiVersionHeaders.ts
```diff
@@ -56,3 +56,34 @@ export default function checkAdminApiVersionHeaders(
...
+  const patchRequestHeaderValue = 'application/merge-patch+json'
+  res.header('Accept-Patch', patchRequestHeaderValue)
+
+  const patchRequestHeader = req.header('Content-Type')
+
+  if (!patchRequestHeader || patchRequestHeader !== patchRequestHeaderValue) {
```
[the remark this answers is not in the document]
FlorentSG: It's updated

src/middlewares/checkAdminApiVersionHeaders.ts
```diff
@@ -56,3 +56,34 @@ export default function checkAdminApiVersionHeaders(
...
+    throw new ParseError(
+      `A 'Content-Type' header is required in the request, of the form 'Content-Type: ${patchRequestHeaderValue}'.`,
```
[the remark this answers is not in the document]
FlorentSG: I've updated the error message and I actually throw a `HeaderError` now which is a 415 `Unsupported Media Type` HTTP error.
I think it's a nicer way to handle those errors.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Patch
Please, let me know what you think.

src/middlewares/users.ts
```diff
@@ -2,7 +2,12 @@ import HttpStatus from '@xpring-eng/http-status'
 import { Request, Response, NextFunction } from 'express'
 
 import getAllAddressInfoFromDatabase from '../data-access/payIds'
-import { insertUser, replaceUser, removeUser } from '../data-access/users'
+import {
+  insertUser,
+  replaceUser,
+  removeUser,
+  replacePayId,
```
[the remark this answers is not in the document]
FlorentSG: It's renamed.

src/middlewares/users.ts
```diff
@@ -214,3 +219,80 @@ export async function deleteUser(
   res.locals.status = HttpStatus.NoContent
   next()
 }
+
+/**
+ * Updates a PayID only, not the addresses.
+ *
+ * @param req - An Express Request object, holding the PayID.
+ * @param res - An Express Response object.
+ * @param next - An Express next() function.-.
```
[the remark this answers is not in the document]
FlorentSG: It's updated.

src/middlewares/users.ts
```diff
@@ -214,3 +219,80 @@ export async function deleteUser(
...
+    (rawOldPayId.match(/\$/gu) || []).length !== 1 ||
+    (rawNewPotentialPayId.match(/\$/gu) || []).length !== 1
+  ) {
+    throw new ParseError(
+      'Bad input. PayIDs must contain only one "$"',
+      ParseErrorType.InvalidPayId,
+    )
+  }
+  const newPayId = rawNewPotentialPayId.toLowerCase()
+  const oldPayId = rawOldPayId.toLowerCase()
```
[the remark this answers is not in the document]
FlorentSG: I agree...
It may be better to focus on this validation first before implementing new endpoints, don't you think?

src/middlewares/users.ts
```diff
@@ -214,3 +219,80 @@ export async function deleteUser(
...
+
+  // Check if the old and new PayID are the same
+  if (oldPayId === newPayId) {
+    throw new ParseError(
+      'The new PayID is the same as the one you are trying to update.',
+      ParseErrorType.InvalidPayId,
+    )
+  }
```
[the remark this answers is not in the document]
FlorentSG: It is not an actual error. It works if we update a PayID to the same value.
I thought it would be good to let the user know he is updating to the same value but maybe that's not necessary.
I removed that part throwing the error and updated the E2E testing.

src/middlewares/users.ts
```diff
@@ -214,3 +219,80 @@ export async function deleteUser(
...
+
+  const account = await replacePayId(oldPayId, newPayId)
+
+  if (account.length === 1) {
+    res.locals.status = HttpStatus.Created
```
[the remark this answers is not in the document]
FlorentSG: I changed that part as we don't retrieve an Account array anymore.
I added a comment to explain why the `Account` object can be null.

src/middlewares/users.ts
```diff
@@ -214,3 +219,80 @@ export async function deleteUser(
...
+    res.locals.payId = account[0].payId
```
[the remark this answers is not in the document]
FlorentSG: Yes, I've updated it

src/routes/adminApiRouter.ts
```diff
@@ -1,9 +1,17 @@
 import * as express from 'express'
 
-import checkAdminApiVersionHeaders from '../middlewares/checkAdminApiVersionHeaders'
+import checkAdminApiVersionHeaders, {
+  checkAdminPatchApiHeaders,
+} from '../middlewares/checkAdminApiVersionHeaders'
```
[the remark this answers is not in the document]
FlorentSG: Yes, I think it makes more sense to name it `checkAdminApiHeaders` now. I wanted to check with you, glad you have the same opinion.
I also removed the default export.

src/routes/adminApiRouter.ts
```diff
@@ -35,6 +43,16 @@ adminApiRouter
   // Delete user route
   .delete('/*', checkAdminApiVersionHeaders, wrapAsync(deleteUser), sendSuccess)
 
+  // Patch user PayID route
+  .patch(
+    '/:payId',
+    express.json({ type: 'application/*+json' }),
```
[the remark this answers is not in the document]
FlorentSG: At this stage, I think we can be explicit and say `application/merge-patch+json`.
I can't think of any downsides by writing this. I've updated the code.

src/routes/adminApiRouter.ts
```diff
@@ -35,6 +43,16 @@ adminApiRouter
...
+    checkAdminApiVersionHeaders,
+    checkAdminPatchApiHeaders,
```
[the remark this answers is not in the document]
FlorentSG: it's renamed

src/middlewares/users.ts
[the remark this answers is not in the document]
hbergren: > It may be better to focus on this validation first before implementing new endpoints, don't you think?

One thing that may be coming is a separate `payid-utility` library that is a standalone library that both Xpring SDK and the `payid` server can take a dependency on.

Something we've heard in the community is that it's too bad that there's no solid parsing / validation of PayIDs and address payloads. But I think the right way to solve that is to put that stuff in a standalone library that people can take a dependency on that, rather than solving it in just the `payid` repo.

So for now, I think implementing PATCH is still valuable, because we can solve the parsing / validation problem down the road.

src/data-access/users.ts
```diff
@@ -41,6 +41,29 @@ export async function insertUser(
...
+): Promise<Account | null> {
+  const account = await knex<Account>('account')
+    .where('payId', oldPayId)
+    .update({ payId: newPayId })
+    .returning('*')
+
+  if (account.length === 1) {
+    return account[0]
+  }
+  return null
```
hbergren, line 64: nit: space.

```suggestion
  }

  return null
```

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -56,3 +61,38 @@ export default function checkAdminApiVersionHeaders(
...
+export function checkAdminApiPatchHeaders(
+  req: Request,
+  res: Response,
+  next: NextFunction,
+): void {
+  // The merge patch format is primarily intended for use with the HTTP PATCH method
+  // as a means of describing a set of modifications to a target resource’s content.
+  // application/merge-patch+json is a Type Specific Variation of the "application/merge-patch" Media Type that uses a
+  // JSON data structure to describe the changes to be made to a target resource.
+  const patchContentType = 'application/merge-patch+json'
```
hbergren, line 83: Great comments. Thanks!

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -56,3 +61,38 @@ export default function checkAdminApiVersionHeaders(
...
+
+  // Add this header to all successful responses. We add it early so even errors will respond with Accept-Patch header.
+  // The Accept-Patch response HTTP header advertises which media-type the server
+  // is able to understand while receiving a PATCH request.
+  res.header('Accept-Patch', patchContentType)
+
+  if (req.header('Content-Type') !== patchContentType) {
+    throw new HeaderError(
```
hbergren, line 91: I love the idea of a `415` custom error. Can we rename it to `ContentTypeError`?

I feel like that is more descriptive. For example, not including the correct `PayID-Version` header is a "header error", but it's not a `415` code error.

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -56,3 +61,38 @@ export default function checkAdminApiVersionHeaders(
...
+      `A specific 'Content-Type' header is required for a PATCH request: 'Content-Type: ${patchContentType}'.`,
+      HeaderErrorType.UnsupportedMediaTypeHeader,
```
hbergren, line 93: You might be able to add this in the base `HeaderError` constructor. Are there multiple `HeaderErrorType`s that you anticipate?

src/utils/errors/headerError.ts
```diff
@@ -0,0 +1,29 @@
...
+  public readonly kind: HeaderErrorType
+
+  /**
+   * The constructor for new HeaderError.
+   *
+   * @param message - An error message.
+   * @param kind - The kind of HeaderError for this error instance.
+   */
+  public constructor(message: string, kind: HeaderErrorType) {
+    // All parsing errors are the result of a 415 Unsupported Media Type error
```
hbergren, line 25:
```suggestion
    // All content type errors are the result of a 415 Unsupported Media Type error
```

test/integration/e2e/admin-api/patchPayId.test.ts
```diff
@@ -0,0 +1,281 @@
+import HttpStatus from '@xpring-eng/http-status'
+import * as request from 'supertest'
+import 'mocha'
```
hbergren, line 3: nit: Have the mocha import up top by itself.

```suggestion
import 'mocha'

import HttpStatus from '@xpring-eng/http-status'
import * as request from 'supertest'
```

test/integration/e2e/admin-api/patchPayId.test.ts
```diff
@@ -0,0 +1,281 @@
...
+  })
+
+  it('Returns a 201 when updating a user PayID', function (done): void {
+    // GIVEN a PayID known to resolve to an account on the PayID service
+    const payId = 'alice$xpring.money'
+    const newPayId = {
+      payId: 'john$xpring.money',
+    }
+
+    // WHEN we make a PATCH request to /users/ with the new PayID to update
```
hbergren, line 25:
```suggestion
    // WHEN we make a PATCH request to /users with the new PayID to update
```

test/integration/e2e/admin-api/patchPayId.test.ts
```diff
@@ -0,0 +1,281 @@
...
+    request(app.adminApiExpress)
+      .patch(`/users/${payId}`)
+      .set('PayID-API-Version', payIdApiVersion)
+      .set('Content-Type', contentType)
+      .send(newPayId)
+      .expect('Content-Type', /text\/plain/u)
```
hbergren, line 31: Why do we expect this content type?

test/integration/e2e/admin-api/patchPayId.test.ts
```diff
@@ -0,0 +1,281 @@
...
+      .expect('Accept-Patch', contentType)
+      // THEN we expect the Location header to be set to the path of the created user resource
+      .expect('Location', `/users/${newPayId.payId}`)
+      // THEN we expect back a 201-Created
```
hbergren, line 35: nit:

```suggestion
      // AND we expect back a 201-Created
```

test/integration/e2e/admin-api/patchPayId.test.ts
```diff
@@ -0,0 +1,281 @@
...
+      .expect(HttpStatus.Created, done)
+  })
+
+  it('Returns a 201 when updating a user PayID with an uppercase PayID', function (done): void {
```
hbergren, line 39: Good test!

test/integration/e2e/admin-api/patchPayId.test.ts
```diff
@@ -0,0 +1,281 @@
...
+
+    // WHEN we make a PATCH request to /users/:payId with the new PayID to update
+    request(app.adminApiExpress)
+      .patch(`/users/${payId}`)
+      .set('PayID-API-Version', payIdApiVersion)
+      .set('Content-Type', contentType)
+      .send(newPayId)
+      .expect('Content-Type', /json/u)
+      .expect('Accept-Patch', contentType)
+      // THEN we expect back a 409 - CONFLICT and our expected error response
```
hbergren, line 190:
```suggestion
      // THEN we expect back a 404 - NOT FOUND and our expected error response
```

test/integration/e2e/admin-api/patchPayId.test.ts
```diff
@@ -0,0 +1,281 @@
...
+      error: 'Unsupported Media Type',
+      message:
+        "A specific 'Content-Type' header is required for a PATCH request: 'Content-Type: application/merge-patch+json'.",
+    }
+
+    // WHEN we make a PATCH request to /users/:payId with the new PayID to update
+    request(app.adminApiExpress)
+      .patch(`/users/${payId}`)
+      .set('PayID-API-Version', payIdApiVersion)
+      // WITHOUT a Content-Type header
```
hbergren, line 241: Great comment.

test/integration/e2e/admin-api/patchPayId.test.ts
```diff
@@ -0,0 +1,281 @@
...
+      .expect('Content-Type', /json/u)
+      .expect('Accept-Patch', contentType)
+      // THEN we expect back a 415 - Unsupported Media Type and our expected error response
+      .expect(HttpStatus.UnsupportedMediaType, expectedErrorResponse, done)
+  })
+
+  it('Returns a 415 - Unsupported Media Type when setting the Content-Type header to a wrong value', function (done): void {
+    // GIVEN a PayID known to resolve to an account on the PayID service
+    const payId = 'alice$127.0.0.1'
+    // AND a request to update that PayID to one known to already exist on the PayID Service
```
hbergren, line 252: Does `chloe` already exist on the PayID service?

hbergren:
Also, here is why I think we shouldn't tackle parsing / validation right now:
https://github.com/payid-org/payid/pull/586#discussion_r451506400

test/integration/e2e/admin-api/patchPayId.test.ts
FlorentSG: Oops, no it doesn't, I corrected that.

test/integration/e2e/admin-api/patchPayId.test.ts
FlorentSG: I actually updated it to `/users/:payId` to match what I wrote in the other tests.

src/middlewares/checkAdminApiHeaders.ts
FlorentSG: Sure, I renamed the class to `ContentTypeError` and the file to `contentTypeError`

src/middlewares/checkAdminApiHeaders.ts
FlorentSG: After searching on the web, I don't think we will need another `HeaderErrorTypes` so I moved it to the constructor directly.

I changed the constructor by including the possibility to throw an error for requests omitting `Content-Type: application/json`. This will be useful for the `POST` and `PUT` endpoints.
I did a test using Postman and at the moment if you try to create a PayID with the correct payload (including a `payId` field) but without `Content-Type`, you will see the following `400` error:
```
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "A `payId` must be provided in the request body."
}
```
which is not the best in my opinion. We should return a `415` in this case. RippleNet also returns a `415` in such cases.
So now with the updated `checkAdminApiContentTypeHeaders` function, it will be possible to do that check.

What do you think?

test/integration/e2e/admin-api/patchPayId.test.ts
FlorentSG: As we don't return a JSON in the response, only a `201 Created`, we don't need `application/json`.
This is also the same when we create a user using the `POST` endpoint, we expect `text/plain`.
Please let me know if I am missing something.

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -1,7 +1,12 @@
 import { Request, Response, NextFunction } from 'express'
 
 import config, { adminApiVersions } from '../config'
-import { ParseError, ParseErrorType } from '../utils/errors'
+import {
+  ParseError,
+  ParseErrorType,
+  HeaderError,
```
hbergren, line 7: This should be `ContentTypeError` now, right?

src/middlewares/checkAdminApiHeaders.ts
hbergren: That's a great observation! Would you make a GitHub issue that captures that we should throw a 415 for POST and PUT when given an invalid Content-Type header?

That can be a follow-on PR after this gets merged.

src/utils/errors/headerError.ts
```diff
@@ -0,0 +1,29 @@
...
+ * An enum containing the different kinds of HeaderError.
+ */
+export enum HeaderErrorType {
+  UnsupportedMediaTypeHeader = 'UnsupportedMediaTypeHeader',
+}
+
+/**
+ * A custom error type to organize logic around 415 - Unsupported Media Type errors.
+ */
+export default class HeaderError extends PayIDError {
```
hbergren, line 15: Hey @FlorentSG , did you push up the code changes where you renamed this to `ContentTypeError` and renamed this file?

FlorentSG:
Hi @hbergren , I was about to push the code when you answered :)
It is pushed now so you should be able to see the updates.

>That's a great observation! Would you make a GitHub issue that captures that we should throw a 415 for POST and PUT when given an invalid Content-Type header?
Yes, I will create it.

>That can be a follow-on PR after this gets merged.
Yes, that was my plan.

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -56,3 +61,44 @@ export default function checkAdminApiVersionHeaders(
...
+     * as a means of describing a set of modifications to a target resource’s content.
+     * Application/merge-patch+json is a Type Specific Variation of the "application/merge-patch" Media Type that uses a
+     * JSON data structure to describe the changes to be made to a target resource.
+     */
+    mediaType = MediaType.ApplicationMergePatchJson
+
+    // Add this header to all successful responses. We add it early so even errors will respond with Accept-Patch header.
+    // The Accept-Patch response HTTP header advertises which media-type the server
+    // is able to understand while receiving a PATCH request.
+    res.header('Accept-Patch', MediaType.ApplicationMergePatchJson)
```
dino-rodriguez, line 96: This header will get added to any response coming from the endpoint, even if it is not a patch request, right ?

test/integration/e2e/admin-api/patchPayId.test.ts
```diff
@@ -0,0 +1,282 @@
...
+      statusCode: 400,
+    }
+
+    // WHEN we make a PATCH request to /users/:payId with the new PayID to update
+    request(app.adminApiExpress)
+      .patch(`/users/${payId}`)
+      .set('PayID-API-Version', payIdApiVersion)
+      .set('Content-Type', contentType)
+      .send(newPayId)
+      .expect('Content-Type', /json/u)
```
dino-rodriguez, line 106: Is this the same as `application/json`?

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -56,3 +61,44 @@ export default function checkAdminApiVersionHeaders(
...
+  }
+
+  if (req.header('Content-Type') !== mediaType) {
```
dino-rodriguez, line 99: One thing I'm not sure I fully understand : Why do we include this merge media type for the `Content-Type` header, but in the response return the regular `application/json` ( or whatever the relevant header would be ) ?

src/middlewares/checkAdminApiHeaders.ts
hbergren: Yeah, but it needs to. This is how the server advertises that:
- It can handle a PATCH request (at least to this endpoint)
- The `Content-Type` that all PATCH requests need to be in.

src/middlewares/checkAdminApiHeaders.ts
hbergren: https://tools.ietf.org/id/draft-snell-merge-patch-02.html

Basically `merge-patch+json` indicates that: "My request includes a sparse JSON object, that is not reflective of the entire resource I am updating. I want you to take my request object, and update the existing resource to have those keys/values, while leaving alone any keys that I do not specify."

So:

```
PATCH `/users/hansbergren$xpring.money
{
  payId: hbergren$xpring.money
}
```

Says: "Update my PayID to `hbergren`". If this were a PUT request, this would also _wipe all my addresses_, but because this is a PATCH with a `merge-patch` ContentType, it leaves my addresses in place.

The alternative to `merge-patch` is `json-patch`, which we decided against.

src/utils/errors/contentTypeError.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export default class ContentTypeError extends PayIDError {
+  public readonly kind: string
+
+  /**
+   * The constructor for new ContentTypeError.
+   *
+   * @param mediaType - The Media Type (also called Content Type) required by the request.
+   */
+  public constructor(mediaType: MediaType) {
```
hbergren, line 25: Can we just make this:

```suggestion
  public constructor(mediaType: 'application/json' | 'application/merge-patch+json') {
```

src/utils/errors/contentTypeError.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import PayIDError from './payIdError'
+
+/**
+ * An enum containing the different kinds of Media Types (but also sometimes called "Content Types") used by the PayID server.
+ */
+export enum MediaType {
+  ApplicationJson = 'application/json',
+  ApplicationMergePatchJson = 'application/merge-patch+json',
+}
```
hbergren, line 11: This is 80% my opinion/intuition/gut, but I don't think `enum` is the right tool to use here.

I don't have a strong argument, but `application/json` and `merge-patch` are both internet standard things, that you can look up on MDN. I think `enum`s are better suited to custom enumerations, (like how we enumerate the different error types for our custom errors).

Also, `MediaType.ApplicationMergePatchJson` is more characters than `application/merge-patch+json`. (A weak argument I know).

Finally, we aren't really using the `MediaType` enum as a union type of its members, except for in the ContentTypeError constructor below. For POST and PUT the content type should _always_ be `application/json` and for PATCH, the content type should _always_ be `application/merge-patch+json`.

If we had more than 2 ContentTypes supported in requests, I could see the argument for turning that into an enumeration, but for right now, I'd prefer if we just hardcoded `application/json` and `application/merge-patch+json`.

Does that make sense? Do you strongly disagree? (Usually I'm able to give a stronger case as to why my opinions are valid, but this is mostly a gut feeling).

test/integration/e2e/admin-api/patchPayId.test.ts
FlorentSG: That's a regex expression so I suppose a string containing `json` would work even if it's not `application/json`. I didn't have time to check today but I can do that on Monday.
I actually used what's currently used already in the tests for the POST or PUT requests.

src/middlewares/checkAdminApiHeaders.ts
FlorentSG: As it's in the `if (req.method === 'PATCH')` block, the `Accept-Patch` header will not be added if it's a POST or PUT method to the same endpoint.
Is that ok?

src/utils/errors/contentTypeError.ts
FlorentSG: At this point, I agree and I am fine with the changes.

We might need another Media Type in the future which is `application/x-www-form-urlencoded` if we ever want PayID to support [OAuth2](https://oauth.net/2/) authentication (that would be for a POST request). This is what we use in RippleNet to get a token and authorize requests to any API endpoint.
Do you think OAuth2 could be used for the Admin APIs one day?

After the changes I had to create a custom type in `checkAdminApiHeaders.ts`:
`  type Mime = 'application/json' | 'application/merge-patch+json'`
There was a conflict otherwise:
```
Argument of type 'string' is not assignable to parameter of type '"application/json" | "application/merge-patch+json"'.ts(2345)
```

image

src/utils/errors/contentTypeError.ts
FlorentSG: I updated the code to reflect your suggestion.

src/utils/errors/contentTypeError.ts
hbergren: > Do you think OAuth2 could be used for the Admin APIs one day?

Absolutely, but that's a ways down the road if we ever get to it.

hbergren: > After the changes I had to create a custom type in checkAdminApiHeaders.ts:
type Mime = 'application/json' | 'application/merge-patch+json'
There was a conflict otherwise:

I bet you could do `const mediaType = 'application/json' as const`

That tells the TypeScript server to type is as the string `application/json` instead of typing it as `string`.

src/middlewares/checkAdminApiHeaders.ts
hbergren: I think the server should always advertise `Accept-Patch`. According to MDN:

> Accept-Patch in response to any method means that PATCH is allowed on the resource identified by the Request-URI. Two common cases lead to this:

So, even a Admin API GET or DELETE request should include the `Accept-Patch` header.

hbergren:
Left a comment here: https://github.com/payid-org/payid/pull/586#discussion_r453085364

src/middlewares/checkAdminApiHeaders.ts
FlorentSG: @hbergren thank you, I've updated the code to add the `Accept-Patch` header to the responses for the requests matching the `/users/:payId` endpoint as explained in MDN.

test/integration/e2e/admin-api/patchPayId.test.ts
FlorentSG: Hi @dino-rodriguez , I checked and it is the same as `application/json`. This is what's shown in https://github.com/visionmedia/supertest also.

src/middlewares/users.ts
```diff
@@ -214,3 +219,73 @@ export async function deleteUser(
...
+  // If we try to update a PayID which doesn't exist, the 'account' object will be null.
+  if (account) {
+    res.locals.status = HttpStatus.Created
+    res.locals.payId = newPayId
+  } else {
+    throw new LookupError(
+      `The PayID ${oldPayId} doesn't exist.`,
+      LookupErrorType.MissingPayId,
+    )
+  }
```
hbergren, line 288: Can we change this to something like:

```
if (!account) {
  throw new LookupError(...)
}

res.locals.status = HttpStatus.Created
...
```

That would kill the `else` statement block, which makes the code a bit easier to understand.

hbergren:
Hey @FlorentSG , sorry for the delayed review. There's one or two more things I want to change before we merge this in, but I would love to merge this this week if possible.

src/utils/errors/contentTypeError.ts
```diff
@@ -0,0 +1,25 @@
+import HttpStatus from '@xpring-eng/http-status'
+
+import PayIDError from './payIdError'
+
+/**
+ * A custom error type to organize logic around 415 - Unsupported Media Type errors.
+ * Https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415.
```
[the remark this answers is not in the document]
FlorentSG: It's updated

src/routes/adminApiRouter.ts
```diff
@@ -19,6 +28,7 @@ adminApiRouter
     '/',
```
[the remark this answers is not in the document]
FlorentSG: I added the OPTIONS endpoint following your link and I also added two E2E tests. Please let me know what you think.

src/middlewares/users.ts
FlorentSG: Yes, it's updated.

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    res.header('Accept-Patch', 'application/merge-patch+json')
+  }
+
+  // A GET and DELETE requests don't need a "Content-Type: application/json" header
+  if (
+    req.header('Content-Type') !== mediaType &&
+    req.method !== 'GET' &&
+    req.method !== 'DELETE'
+  ) {
```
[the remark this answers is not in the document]
FlorentSG: Yes, that's better, I adjusted it.

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   *  This Regex will match the endpoint /users/:payId BUT will not match any sub endpoints.
+   *
+   * /users/alice$xpring.money - Matches
+   * /users/alicexpring.money - Matches (even if malformed PayID)
+   * /users/alice$xpring.money/addresses - Does not match
+   * /users/alice$xpring.money/addresses/example - Does not match.
+   */
+  // const patchEndpointRegex = /\/users\/[^\/]+$/giu
+  const patchEndpointRegex = /\/users\/[^/]+$/giu
```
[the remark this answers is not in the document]
FlorentSG: @hbergren I am glad you mentioned this as I had the same idea before my last push.
My problem was that Typescript was complaining when I didn't use the `req` or `res` parameter and I didn't know how to tell Typescript to ignore that.
I finally found out that adding an underscore in front of the `req` or `res` parameter solves the issue :)
So now we have two different functions and it's much better and easier to use. 
Please let me know what you think of the two new functions.

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -58,17 +58,16 @@ export function checkAdminApiVersionHeaders(
 }
 
 /**
- * A middleware asserting that all Admin requests have an appropriate Content-Type header and an Accept-Patch header in response.
+ * A middleware asserting that Admin requests have an appropriate Content-Type header.
  *
  * @param req - An Express Request object.
- * @param res - An Express Response object.
+ * @param _res - An Express Response object. Unused parameters can be conventionally underscored to avoid the check.
```
hbergren, line 64:
```suggestion
 * @param _res - An Express Response object.
```

hbergren: This is pretty conventional, I don't think we need an explanation comment here.

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -58,17 +58,16 @@ export function checkAdminApiVersionHeaders(
...
  * @throws A ParseError if the Content-Type header is missing, malformed, or unsupported.
  */
-export function checkAdminApiContentTypeHeaders(
+export function checkContentType(
   req: Request,
-  res: Response,
+  _res: Response,
   next: NextFunction,
 ): void {
   type Mime = 'application/json' | 'application/merge-patch+json'
```
hbergren, line 73:
```suggestion
  type MimeType = 'application/json' | 'application/merge-patch+json'
```

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -85,47 +84,61 @@ export function checkAdminApiContentTypeHeaders(
...
+    ['POST', 'PUT', 'PATCH'].includes(req.method)
   ) {
     throw new ContentTypeError(mediaType)
   }
 
   next()
 }
+
+/**
+ * A middleware asserting that the request has an Accept-Patch header in response.
```
hbergren, line 101: I don't know that we're "asserting" this.

It's more that the middleware is actually putting an `Accept-Patch` header in the response, right?

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -85,47 +84,61 @@ export function checkAdminApiContentTypeHeaders(
...
+ *
+ * @param _req - An Express Request object. Unused parameters can be conventionally underscored to avoid the check.
```
hbergren, line 103:
```suggestion
 * @param _req - An Express Request object.
```

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -85,47 +84,61 @@ export function checkAdminApiContentTypeHeaders(
...
+ * @param res - An Express Response object.
+ * @param next - An Express next() function.
+ */
+export function addAcceptPatchHeader(
+  _req: Request,
+  res: Response,
+  next: NextFunction,
+): void {
+  /**
+   * Add this header to all responses.
```
hbergren, line 113:
```suggestion
   * Add this header to the response.
```

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -85,47 +84,61 @@ export function checkAdminApiContentTypeHeaders(
...
+}
+
+/**
+ * A middleware adding an Allow header in the response.
+ *
+ * @param _req - An Express Request object. Unused parameters can be conventionally underscored to avoid the check.
+ * @param res - An Express Response object.
+ * @param next - An Express next() function.
+ */
+export function addAllowHeader(
```
hbergren, line 129: I don't think we need this function. I'm pretty sure Express automatically generates an `Allow` header for options requests. Can you comment out the places you invoke this function and test that theory?

(It could be that us explicitly adding an `.options()` route broke the auto-generated `Allow` response header, but I'd like to verify that).

src/routes/adminApiRouter.ts
```diff
@@ -38,24 +46,40 @@ adminApiRouter
...
     express.json({ type: 'application/merge-patch+json' }),
     checkAdminApiVersionHeaders,
-    checkAdminApiContentTypeHeaders,
+    addAcceptPatchHeader,
+    checkContentType,
     wrapAsync(patchUserPayId),
     sendSuccess,
   )
 
+  .options(
```
hbergren, line 75: nit: I think `options` should be before GET. There's sort of a semantic ordering of HTTP request types in my mind.

FlorentSG:
Hi @dino-rodriguez , @hbergren , I am trying to push the changes after Hans review but I get the following error:

image

It seems the issue comes from the `offset` in `version.sh` which starts at 1. Shouldn't it start at 0? I don't see a `v` when I execute `git describe --tags` on my Macbook.
```
# Drop the 'v' from the Git tag for the comparison.
if [[ $npm_version != ${git_tag_version:1} ]]; then
```

I have git version 2.20.1 (Apple Git-117)

image

hbergren:
Hey @FlorentSG , @dino-rodriguez has been doing some work on improving our release tooling. I'll have him take a look at this when he starts his day.

test/integration/e2e/admin-api/patchPayId.test.ts
dino-rodriguez: Great, sounds good

dino-rodriguez:
@FlorentSG Hey Florent, could you try pulling down the latest git tags?

I just deleted my PayID repo, and created a fresh one and tested locally, so what is on remote should work. There might be a discrepancy because I've been poking around on the tags over the past few days.

src/middlewares/checkAdminApiHeaders.ts
FlorentSG: Yes, I changed the comment. Thanks.

src/middlewares/checkAdminApiHeaders.ts
FlorentSG: As discussed previously, Express can generate automatically an `OPTIONS` endpoint with the `Allow` header but it is missing `Accept-Patch`.
I looked at this issue which seems to be what I experienced, I tried what was mentioned there but it doesn't work. `Accept-Patch` is still absent.
https://github.com/expressjs/express/issues/2878

So I came back to explicitly adding the `.options()` route.
Please let me know if that's ok.

src/routes/adminApiRouter.ts
FlorentSG: I moved it before GET.

FlorentSG:
Hi @hbergren , @dino-rodriguez , thank you, deleting the local branch `1.0.1` worked and I could push the changes.
Please let me know if you have any other comments.

src/routes/adminApiRouter.ts
```diff
@@ -22,6 +22,15 @@ const adminApiRouter = express.Router()
  * Routes for the PayID Admin API.
  */
 adminApiRouter
+  // Options user PayID route
+  .options(
+    '/:payId',
+    checkAdminApiVersionHeaders,
+    addAcceptPatchHeader,
+    addAllowHeader,
```
hbergren, line 30: Can you change the order of this?

We should have `addAllowHeader` and `addAcceptPatchHeader` before `checkAdminApiVersionHeaders`.

We don't want to send a response without an `Allow` or `Accept-Patch` just because the `PayID-API-Version` header was malformed.

(For this middleware and every other middleware).

FlorentSG: Hi, @hbergren I changed the order and updated the E2E `options` tests to include the `Allow` and `Accept-Patch` headers when it returns a 400.

src/middlewares/checkAdminApiHeaders.ts
```diff
@@ -0,0 +1,144 @@
...
+     *
+     * The merge patch format is primarily intended for use with the HTTP PATCH method
+     * as a means of describing a set of modifications to a target resource’s content.
+     * Application/merge-patch+json is a Type Specific Variation of the "application/merge-patch" Media Type that uses a
+     * JSON data structure to describe the changes to be made to a target resource.
+     */
+    mediaType = 'application/merge-patch+json'
+  }
+
+  // POST, PUT and PATCH requests need a "Content-Type: application/json" header
```
hbergren, line 89:
```suggestion
  // POST, PUT and PATCH requests need a valid Content-Type header
```

test/integration/e2e/admin-api/optionsUsersPayId.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    request(app.adminApiExpress)
+      .options(`/users/${payId}`)
+      // WITHOUT the 'PayID-API-Version' header
+      .expect('Content-Type', /json/u)
+      // THEN we expect the Allow header in the response
+      .expect('Allow', allowHeader)
+      // THEN we expect the Accept-Patch header in the response
+      .expect('Accept-Patch', contentType)
+      // THEN we expect back a 400 - Bad Request
+      .expect(HttpStatus.BadRequest, expectedErrorResponse, done)
```
hbergren, line 63: Great test! Thank you.

src/data-access/users.ts
```diff
@@ -41,6 +41,30 @@ export async function insertUser(
...
+ * @returns The updated user Account.
+ */
+export async function replaceUserPayId(
+  oldPayId: string,
+  newPayId: string,
+): Promise<Account | null> {
+  const account = await knex<Account>('account')
+    .where('payId', oldPayId)
+    .update({ payId: newPayId })
+    .returning('*')
```
austin-king, line 59: nice and concise

src/middlewares/users.ts
```diff
@@ -214,3 +219,73 @@ export async function deleteUser(
...
+): Promise<void> {
+  const rawOldPayId = req.params.payId
+  if (!rawOldPayId) {
+    throw new ParseError(
+      'A `payId` must be provided in the path. A well-formed API call would look like `PATCH /users/alice$xpring.money`.',
+      ParseErrorType.MissingPayId,
+    )
+  }
+
+  // "Potential" because we don't know yet if there will be an error or not
```
austin-king, line 246: Appreciative of the comment

FlorentSG:
Thank you @austin-king for your comment, I agree with what you said and I changed the file name to `adminApiHeaders`.
@hbergren I will let you do the merge if you don't have more comments.
Thanks

PR #594, opened 2020-07-09
Add a report for count of PayIDs

hbergren:

Adds a report for the count of the number of PayIDs.

Also rename the existing report logic to reflect that it is actually a count of addresses.

### Context of Change

We currently only capture a count of addresses grouped by `(paymentNetwork, environment)`, but it can be misleading if you think that is a count of unique PayIDs. Now, we have a separate Prometheus Gauge that counts up the number of unique PayIDs per server.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added some relevant tests and `npm run test` is green.

### DIFF
--- a/src/data-access/reports.ts
+++ b/src/data-access/reports.ts
@@ -1,25 +1,39 @@
 import knex from '../db/knex'
-import { PayIdCount } from '../types/reports'
+import { AddressCount } from '../types/reports'
 
 /**
- * Retrieve count of PayIDs, grouped by payment network and environment.
+ * Retrieve count of addresses, grouped by payment network and environment.
  *
- * @returns A list with the number of PayIDs that have a given (paymentNetwork, environment) tuple,
+ * @returns A list with the number of addresses that have a given (paymentNetwork, environment) tuple,
  *          ordered by (paymentNetwork, environment).
  */
-export default async function getPayIdCounts(): Promise<PayIdCount[]> {
-  const payIdCounts: PayIdCount[] = await knex
+export async function getAddressCounts(): Promise<AddressCount[]> {
+  const addressCounts: AddressCount[] = await knex
     .select('address.paymentNetwork', 'address.environment')
     .count('* as count')
-    .from<PayIdCount>('address')
+    .from<AddressCount>('address')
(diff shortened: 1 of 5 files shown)


src/data-access/reports.ts
```diff
@@ -1,25 +1,39 @@
 import knex from '../db/knex'
-import { PayIdCount } from '../types/reports'
+import { AddressCount } from '../types/reports'
```
hbergren, line 2: Rename this to AddressCount, to reflect that this is actually a report for the count of addresses.

src/services/metrics.ts
```diff
@@ -28,8 +28,11 @@ class Metrics {
     'paymentNetwork' | 'environment' | 'org' | 'result'
   >
 
-  /** Prometheus Gauge for reporting the current count of PayIDs by [network, environment, org]. */
-  private readonly payIdGauge: Gauge<'paymentNetwork' | 'environment' | 'org'>
+  /** Prometheus Gauge for reporting the current count of addresses by [network, environment, org]. */
+  private readonly addressGauge: Gauge<'paymentNetwork' | 'environment' | 'org'>
+
+  /** Prometheus Gauge for reporting the current count of PayIDs by org. */
+  private readonly payIdGauge: Gauge<'org'>
```
hbergren, line 35: We need a separate Gauge, but we can reuse the existing `payIdGaugeRegistry`.

src/services/metrics.ts
```diff
@@ -48,12 +51,21 @@ class Metrics {
       registers: [this.payIdLookupCounterRegistry],
     })
 
-    this.payIdGauge = new Gauge({
+    this.addressGauge = new Gauge({
+      // This should really be address_count, but changing it now would break metrics,
```
dino-rodriguez, line 55: Can we make this a TODO comment if we are unhappy with its current state ?

hbergren: It's hard to make this a TODO... because it would be a breaking change. I don't think we would ever actually change this until:
- We have a REST API
- We switch from Prometheus to something else.

Issue #595, opened 2020-07-09
DELETE does not actually delete a user?


hbergren:
## Expected Behavior

When I DELETE a user using the Admin API, and then subsequently GET that user using the public API with a header of `payid+json`, I should get a 404 - Not Found.

## Actual Behavior

I get a 200 - OK, with a payload including the `payId` and an empty addresses array.

## Context

When I inspect my local database, this PayID does _not actually exist_ anymore after the DELETE operation, so there is probably a bug in our Public API and how we handle the `payid+json` header.

## Steps to Reproduce

1. Spin up the PayID server
2. `DELETE 127.0.0.1:8081/users/alice$127.0.0.1`
3. `GET http://127.0.0.1:8080/alice` + `Accept: application/payid+json`

hbergren:
Closed by https://github.com/payid-org/payid/pull/598

PR #598, opened 2020-07-14
Fix payid+json incorrect 200 error

hbergren:

Fix bug in header processing logic, and remove needlessly complex logic for the "Not Found" message we provide.

### Context of Change

There was a bug in our header processing logic, which, if you requested `payid+json` to a non-existent PayID would respond with a `200 - OK`, claiming that PayID existed with no addresses, rather than responding with a `404 - Not Found`.

Relevant issue: https://github.com/payid-org/payid/issues/595

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added a test for the `payid+json` to a nonexistent PayID case, and it now passes (would fail on `master`).

### DIFF
--- a/src/middlewares/payIds.ts
+++ b/src/middlewares/payIds.ts
@@ -53,17 +53,7 @@ export default async function getPaymentInfo(
   // TODO:(hbergren) Distinguish between missing PayID in system, and missing address for paymentNetwork/environment.
   // Respond with a 404 if we can't find the requested payment information
   if (preferredAddressInfo === undefined) {
-    let message = `Payment information for ${payId} could not be found.`
-    // When we only have a single accept type, we can give a more detailed error message
-    if (parsedAcceptHeaders.length === 1) {
-      const { paymentNetwork, environment } = parsedAcceptHeaders[0]
-      message = `Payment information for ${payId} in ${paymentNetwork} `
-      // eslint-disable-next-line max-depth -- TODO:(@dino-rodriguez) This should be refactored
-      if (environment) {
-        message += `on ${environment} `
-      }
-      message += 'could not be found.'
-    }
+    // Record metrics for 404s
     parsedAcceptHeaders.forEach((acceptType) =>
       metrics.recordPayIdLookupResult(
         false,
@@ -72,7 +62,10 @@ export default async function getPaymentInfo(
       ),
     )
 
-    throw new LookupError(message, LookupErrorType.Unknown)
+    throw new LookupError(
+      `Payment information for ${payId} could not be found.`,
+      LookupErrorType.Unknown,
+    )
   }
 
   const {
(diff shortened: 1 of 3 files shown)


src/middlewares/payIds.ts
```diff
@@ -53,17 +53,7 @@ export default async function getPaymentInfo(
...
-    // When we only have a single accept type, we can give a more detailed error message
-    if (parsedAcceptHeaders.length === 1) {
-      const { paymentNetwork, environment } = parsedAcceptHeaders[0]
-      message = `Payment information for ${payId} in ${paymentNetwork} `
-      // eslint-disable-next-line max-depth -- TODO:(@dino-rodriguez) This should be refactored
-      if (environment) {
-        message += `on ${environment} `
-      }
-      message += 'could not be found.'
-    }
```
hbergren, line 66: This felt complicated for no real reason, so I just cut it.

src/services/basePayId.ts
```diff
@@ -57,6 +57,10 @@ export function getPreferredAddressHeaderPair(
       preferredAddresses: readonly AddressInformation[]
     }
   | undefined {
+  if (allAddresses.length === 0) {
+    return undefined
+  }
```
hbergren, line 62: This was the bug. By not returning `undefined` when we had no addresses, we would fail our `if(x === undefined)` check in the middleware, and not 404 correctly.

test/integration/e2e/public-api/basePayId.test.ts
```diff
@@ -147,8 +147,7 @@ describe('E2E - publicAPIRouter - Base PayID', function (): void {
     const expectedErrorResponse = {
       statusCode: 404,
       error: 'Not Found',
-      message:
-        'Payment information for johndoe$127.0.0.1 in XRPL on TESTNET could not be found.',
+      message: 'Payment information for johndoe$127.0.0.1 could not be found.',
```
hbergren, line 150: Update random test messages because I changed the 404 not found message.

test/integration/e2e/public-api/basePayId.test.ts
```diff
@@ -182,6 +180,26 @@ describe('E2E - publicAPIRouter - Base PayID', function (): void {
...
+
+    // WHEN we request all addresses for that PayID
+    request(app.publicApiExpress)
+      .get(payId)
+      .set('PayID-Version', '1.0')
+      .set('Accept', acceptHeader)
+      // THEN we get back a 404 with the expected error response.
+      .expect(HttpStatus.NotFound, expectedErrorResponse, done)
+    // })
+  })
```
hbergren, line 201: This is the test that verifies we no longer have a bug in our header processing logic.

src/services/basePayId.ts
dino-rodriguez: Awesome, nice catch.

src/middlewares/payIds.ts
dino-rodriguez: Definitely a little much for the error message.

PR #601, opened 2020-07-15
refactor: config as constructor parameter

dino-rodriguez:

Configuration is passed through a constructor parameter.

Database access functions are also passed in through a constructor parameter.

### Context of Change

We need to do this to be able to separate metrics from the PayID server.

This decouples the metrics configuration & the database access functions from the core metrics library.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Tests passing


src/services/metrics.ts
```diff
@@ -2,8 +2,9 @@ import { hostname } from 'os'
 
 import { Counter, Gauge, Pushgateway, Registry } from 'prom-client'
```
dino-rodriguez, line 4: I'm getting an over >250 line length warning. @hbergren any idea how to disable this in the file? 

When I do `eslint-disable @typescript-eslint/no-explicit-any -- Insert message`

I get a lint error : `Definition for rule '@typescript-eslint/max-lines' was not found`

src/services/metrics.ts
```diff
@@ -2,8 +2,9 @@ import { hostname } from 'os'
...
 
-import config from '../config'
+import configuration from '../config'
```
hbergren, line 5: nit: Any reason why you changed this from `config` to `configuration`?

I know it doesn't matter with default exports, but the actual object we export from `../config` is named `config`, so it seems reasonable to be consistent.

src/services/metrics.ts
hbergren: > /* eslint-disable max-lines -- I wrote too much code */

It's not a rule under `@typescript-eslint`, it's a core ESLint rule.

I wonder if we can break this file up into some smaller helper files when we turn it into a library so you don't have to disable this rule.

src/services/metrics.ts
```diff
@@ -39,8 +40,27 @@ class Metrics {
   private recurringMetricsPushTimeout?: NodeJS.Timeout
   private recurringMetricsTimeout?: NodeJS.Timeout
 
-  // TODO:(hbergren) Should config be a parameter here?
-  public constructor() {
+  // These are passed in from the constructor
+  // This will allow us to extract this into a separate library
+  private readonly config: typeof configuration.metrics
+  private readonly getAddressCounts: () => Promise<AddressCount[]>
+  private readonly getPayIdCount: () => Promise<number>
```
hbergren, line 47: Clever!

src/services/metrics.ts
```diff
@@ -39,8 +40,27 @@ class Metrics {
...
+
+  /**
+   * Create a new Metrics instance.
+   *
+   * @param config - The metrics configuration object.
+   * @param addressCountFn - A function to retrieve count of addresses, grouped by payment network and environment.
+   * @param payIdCountFn - A function to retrieve the count of PayIds in the database.
```
hbergren, line 54:
```suggestion
   * @param payIdCountFn - A function to retrieve the count of PayIDs in the database.
```

src/services/metrics.ts
dino-rodriguez: Yeah, I think we will be able to break this up. There is one function that is outside of the class that's pretty long.

Breaking that out into it's own file should resolve.

Cool, that rule worked, thanks

src/services/metrics.ts
dino-rodriguez: I changed it because using `config` as a class variable as well was causing a name conflict.

I opted to change the import name since it had less references, instead of changing the class variable, which would mean I would use the longer `configuration` in more places.

src/services/metrics.ts
dino-rodriguez: Thanks!

src/services/metrics.ts
```diff
@@ -1,9 +1,11 @@
+/* eslint-disable max-lines -- Will be fixed in standalone metrics lib. */
```
hbergren, line 1: 👍 Good comment 😄

dino-rodriguez:
@austin-king

Yeah, good question. So basically, the main changes are generating new class variables from a constructor :

```
private readonly config: typeof configuration.metrics
private readonly getAddressCounts: () => Promise<AddressCount[]>
private readonly getPayIdCount: () => Promise<number>
```

The code in the library is basically going to be this code ripped out. The reason we need to pass those 3 things into the constructor are because they will not be bundled in the library.

- Configuration will live in the PayID server, or some other separate file, therefore we should load it in via the constructor so the library can accept any configuration
- The other two functions are data access functions -- similarly these won't live in the library ( the library won't do any database stuff ), so they should be passed in by the PayID server ( or if somebody rolled their own implementation, their own custom database access functions that adhere to these signatures )

The reason I'm doing the work here is to make sure that my refactor works ( tests continue to pass, etc. )

With everything passing, then it should be trivial for me to move this code into the library, and use the library directly

PR #603, opened 2020-07-16
refactor+docs: release tooling

dino-rodriguez:

Various refactors to simplify our release scripts.

A guide on how to cut a release.

### Context of Change

Since we squash commits, pushing tags to a branch won't work, because once the
branch is merged to master, a new commit is made, and the tag reference is lost.

After looking further, I don't think any of the GitHub merge options would work for
releases.

I thought rebase & merge might, but according to the GitHub docs : "Rebase and merge on
GitHub will always update the committer information and create new commit SHAs,
whereas git rebase outside of GitHub does not change the committer information"

One way to solve this problem is push directly to master for cutting new releases.

To achieve this, I initially disabled branch protection rules on master for
administrators.

However, Hans indicated that this is not a best practice, and suggested using CI instead of pushing to master. Until those changes are implemented, the process will be a little more lengthy, so I've included detailed instructions on how to go from 0 to release.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Before :

- Scripts in one file
- No `cut` script

After :
- Scripts in separate files, no need to invoke bash
- Cut script

## Test Plan

I tested this locally.


dino-rodriguez:
@austin-king `npm run compareVersions` will run automatically when you run bump, so no need to worry about that.

I'm going to include a README with the suggested workflow.

package.json
```diff
@@ -31,8 +31,7 @@
     "devEnvUp": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose up -d",
     "devDbUp": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose up -d db",
     "devDown": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose down",
-    "compareVersions": "bash -c '. ./scripts/version.sh && compare_versions $npm_package_version'",
```
dino-rodriguez, line 34: For now I only need to run this in bump, so I dropped it as an NPM script. The user should never have to run this directly.

package.json
```diff
@@ -20,8 +20,8 @@
     "copyHTML": "copyfiles -u 1 src/html/* build/src",
     "copySQL": "copyfiles -u 1 **/*.sql build/src",
     "copyJSON": "copyfiles -u 1 src/**/*.json build/src",
-    "lint": "eslint . --ext .ts --fix --max-warnings 0 && prettier --write '**/*.{md,json}' && npm run compareVersions",
```
dino-rodriguez, line 23: This was causing issues because some people didn't have all tags locally, so I removed it. Running this check on `bump` should be sufficient since the versions should only change with this process now.

package.json
```diff
@@ -31,8 +31,7 @@
     "devEnvUp": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose up -d",
     "devDbUp": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose up -d db",
     "devDown": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose down",
-    "compareVersions": "bash -c '. ./scripts/version.sh && compare_versions $npm_package_version'",
-    "bump": "bash -c '. ./scripts/version.sh && bump $npm_package_version $1' -l ",
```
dino-rodriguez, line 35: I refactored the helper library to be separate from the script themselves. The scripts run in bash, so I no longer need to invoke it here.

release.md
```diff
@@ -0,0 +1,41 @@
...
+### release
+
+To create a release draft from the command line, run:
+
+`npm run release` ( you may need to setup `hub` locally )
+
+## Cutting a Release
+
+The full process for cutting a release is as follows:
+
```
dino-rodriguez, line 20: I added this in the PR description, but adding here just in case:

I developed a method that would allow you to run all of this in one click, but it required that branch protections be disabled on master. The reason is that it involved pushing everything directly to master, and so you could do it all without waiting on PR approvals.

After discussing with him, we agreed that this was not an ideal best practice. The solution we came up with is that you need to push up a PR, and then CI would handle the rest.

I need to implement that, so until then, I am providing detailed instructions here on how to do this manually.

scripts/version/bump.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
dino-rodriguez, line 1: Because this is a bash script we no longer need to explicitly invoke this in the `package.json`

scripts/version/version.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Bumps the NPM version, create a git tag, & bump the Docker version.
```
dino-rodriguez, line 3: We aren't bumping the git tag anymore.

Since we squash commits, they each merged PR results in a new commit.

This means that we either have to:

- Do the NPM version bumping and pushing directly on master, so we can push the tags up alongside the `package.json`
- Open a PR and merge it, and then do the tags work

For the reasons outlined above ( best practice, branch restrictions ), we went with number 2, which means we don't want to make tags in the bump script.

scripts/version/version.sh
```diff
@@ -14,35 +14,38 @@ function bump() {
...
    else
-      # Make sure versions are equal
-      compare_versions $current_version
+      declare ir versions_match=$(compare_versions current_version)
 
-      # Bump NPM version and git tag
-      declare -r new_version=$(npm version $bump_increment)
+      # Bump only if Git & NPM versions are equal
+      if [[ versions_match ]]; then
+        declare -r new_version=$(npm --no-git-tag-version version $bump_increment)
```
dino-rodriguez, line 21: Do not bump git tag

scripts/version/version.sh
```diff
@@ -14,35 +14,38 @@ function bump() {
...
 
-      # Bump the Docker image version by finding and replacing the old
-      # version for the new. Drop the 'v' from the Git tag.
-      sed -i "" "s|payid:$current_version|payid:${new_version:1}|g" docker-compose.yml
```
dino-rodriguez, line 25: No more interpolation necessary for Docker compose

scripts/version/version.sh
```diff
@@ -14,35 +14,38 @@ function bump() {
...
+        echo "NPM Version: "${new_version:1}
```
dino-rodriguez, line 23: Abstracting this into the `bump` script.

This error message is useful to the end user, and `compareVersions` is a helper that should return booleans

scripts/version/version.sh
```diff
@@ -14,35 +14,38 @@ function bump() {
...
 
    # Drop the 'v' from the Git tag for the comparison.
    if [[ $npm_version != ${git_tag_version:1} ]]; then
-      error "Version Mismatch:
-
-      NPM Version: "$npm_version"
-      Git Tag Version: "${git_tag_version:1}"
-      "
+     false
    else
```
dino-rodriguez, line 47: Change this to booleans to make it more useful for validation

PR #605, opened 2020-07-20
Add PayID Discovery Endpoint

nkramer44:

* Add new GET resource for `/.well-known/webfinger` to support PayID Discovery
* Add middleware to construct a JRD for Discovery, which imports a static JSON file with a collection of JRD Links
* Enable JSON module resolution in the TS config
* Copy JSON files into the /build directory in the `build` npm goal so that the Discovery Link config file gets transferred to the build

### Context of Change

To enable Easy Checkout in ChefXRP and the GiveDirectly demo that we have, we need to expose a PayID Discovery endpoint (https://github.com/payid-org/rfcs/blob/master/dist/spec/payid-discovery.txt) that serves up a PayID Easy Checkout Link (https://github.com/nkramer44/rfcs/blob/is/payid-easy-checkout/dist/spec/payid-easy-checkout.txt), so that users can automatically get redirected from the merchant demo UIs to the Xpring wallet.

### Type of Change

## Before / After
New endpoint to serve PayID Discovery payload.
New middleware to construct payload.
Newly enabled JSON module resolution in tsconfig

## Test Plan
Because this is such a small change, it doesn't really make sense to pull the core logic into a service to test.  Express modules are terrible to test, so I've just tested this locally and have not written any unit tests.

If there are any big objections to this, please let me know.

## Future Tasks
In `src/middlewares/constructJrd`, I had to disable the `node/file-extension-in-import` eslint rule in order to import the JSON file.  Once this gets merged, I will open a PR in the global linter config to relax this rule.


src/discoveryLinks.json
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "rel": "https://payid.org/ns/payid-easy-checkout/1.0",
```
hbergren, line 3: Should this URL resolve to something instead of `404`?

src/discoveryLinks.json
```diff
@@ -0,0 +1,6 @@
...
+    "template": "https://dev.wallet.xpring.io/wallet/xrp/testnet/payto?receiverPayId={receiverPayId}&amount={amount}&nextUrl={nextUrl}?easyCheckoutSuccess=true"
```
hbergren, line 4: Is this template syntax part of WebFinger? (Should I go read the WebFinger RFC?)

src/middlewares/constructJrd.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @returns A Promise resolving to nothing.
+ * @throws ParseError if the PayID is missing from the request parameters.
+ */
+export default async function constructJrd(
+  req: Request,
+  res: Response,
+  next: NextFunction,
+): Promise<void> {
+  const payId = req.query.resource
+  if (!payId || Array.isArray(payId) || typeof payId !== 'string') {
```
hbergren, line 22: Not sure what the array check adds here. Am I missing something?

```suggestion
  if (!payId || typeof payId !== 'string') {
```

src/routes/publicApiRouter.ts
```diff
@@ -39,6 +40,9 @@ publicApiRouter
   // Health route
   .get('/status/health', sendSuccess)
 
+  // PayID Discovery route
+  .get('/.well-known/webfinger', constructJrd, sendSuccess)
```
hbergren, line 44: So this is independent of an actual PayID request, so we don't need a `PayID-Version` header, correct?

src/discoveryLinks.json
nkramer44: The URL is just an enumeration of what the link can be used for.  Each link relation type will have its own URL, but the URL isn't used for anything except differentiating and identifying link types.  I'm not totally sure why it's typed as a URL, but that's just what the webfinger/JRD spec says

src/discoveryLinks.json
nkramer44: Nope, it's part of the Easy Checkout protocol.  I'd tell you to go read it now, but I believe we'll be getting rid of the template and replacing it with an href

src/middlewares/constructJrd.ts
nkramer44: I think `req.query.resource` is of type `string | string[] | ParsedQs | ParsedQs[]`, so in order to correctly cast it to `string`, I think we needed that check

src/routes/publicApiRouter.ts
nkramer44: Correct.  We can think about versioning at some point, but it should probably be a separate version, like `PayID-Discovery-Version`

src/middlewares/constructJrd.ts
hbergren: Got it. I'm reading over the WebFinger RFC now, and I didn't think through how this worked.

So basically you make a `GET example.com/.well-known/webfinger?resource=hbergren$example.com` request to get the discovery endpoints for my PayID, correct?

And because `resource` is a query parameter, it can either be a string or an array.

Can you add a comment to that effect above this line to aid reader understanding?

src/discoveryLinks.json
hbergren: I found this JRD example: https://www.packetizer.com/json/jrd/

And both of the `rel` URLs resolve to something:
- https://webfinger.net/rel/profile-page/
- https://www.packetizer.com/rel/blog

Does it make sense we would do something like that? I'm also not sure why we use `/ns/` as a path instead of `/rel/` like these examples.

hbergren: It sounds like from https://tools.ietf.org/html/rfc7033#section-8.4, that we _would_ want to have application specifications for our link types, right?

hbergren:
Few thoughts, and feel free to say these are either out of scope or will be implemented in a future PR. I think most of these _are_ out of scope, but are questions I have now that we are introducing a WebFinger endpoint.

1) Should we allow [filtering the WebFinger resource by use of the `rel` query parameter](https://tools.ietf.org/html/rfc7033#section-4.3)?

2) What is the "right way" to build PayID Discovery (not wallet discovery)? I know we've talked about a template thing where you can map `hbergren$example.com` => `example.com/payid/users/hbergren` somehow... but [WebFinger also supports an _aliases_ property](https://tools.ietf.org/html/rfc7033#section-4.4.2).

You could use aliases like this (I think):
```json
{
  "subject": "hbergren$example.com",
  "aliases": ["https://example.com/payid/users/hbergren"],
}
```

3) Keefer and I have discussed how it might be useful if a PayID server advertised its capabilities in some way. Is `.well-known/webfinger` the right place to do that? In the [WebFinger RFC, there is an example where the server advertises its version in the properties array](https://tools.ietf.org/html/rfc7033#section-3.2). Should we have a `properties` field that advertises things like:
  - The `PayID-Version` of the server?
  - Whether or not V. PayID is enabled for the server?
  - Whether or not Travel Rule is enabled for the server (far future)?

src/discoveryLinks.json
nkramer44: > Does it make sense we would do something like that?
I don't see a reason why we couldn't put a UI at those URLs, but my understanding is that those URLs are just enumerations.  Putting a UI in front of them seems premature, given we don't really have anyone using this. Even if they resolve to something, a PayID Discovery client would never actually query the 'rel' URL. 

> I'm also not sure why we use /ns/ as a path instead of /rel/ like these examples.
Honestly not sure where this came from.  The PayID Discovery RFC defines a link relation type of https://payid.org/ns/payid-discovery-url/1.0, so I just went off of that for Easy Checkout.  @sappenin can you clarify?

> It sounds like from https://tools.ietf.org/html/rfc7033#section-8.4, that we would want to have application specifications for our link types, right?
Correct.  We now have two protocols, PayID Discovery and Easy Checkout, which both define their own links

src/middlewares/constructJrd.ts
nkramer44: Yep you got it.  Will do

nkramer44:
>Should we allow filtering the WebFinger resource by use of the rel query parameter?

Maybe in the future.  For now we will have 2, maybe 3 links, so that seems unnecessary.

Re - aliases vs templates: I believe @sappenin decided to only use the `href` and `template` fields.  `href` can basically do whatever `aliases` can do, except that `aliases` is a collection.  I don't see the use case for having multiple aliases for a single PayID, so `href` works fine.  `template` on the other hand allows a PayID Discovery server to return a mostly static response, which is a probable case for something like a vanity PayID where the PayID endpoint is actually something like `https://xpring.io/payid/nkramer44`.  @sappenin Please correct me if I'm wrong.

3. I think that's a great idea.  We've also been thinking about providing account level metadata around supported assets/networks without exposing users address info.  We see that some wallets don't want to expose their users addresses to the world, which is fair, but for Easy Checkout, the merchant doesn't actually need to know the address info of the sender.  It would be nice for a merchant to be able to figure out which networks a user has associated with their PayID to better figure out if the transaction is possible from their wallet.

Overall, I think you're right on most of this stuff.  This PR follows the RFCs we have out currently, although those RFCs could definitely use some love.  I'd suggest a quick read of [PayID Discovery](https://github.com/payid-org/rfcs/blob/master/dist/spec/payid-discovery.txt), and I also have a PR out for Easy Checkout [here](https://github.com/nkramer44/rfcs/pull/1).  Feel free to make comments on the PR or open a PR on the PayID Discovery spec

src/middlewares/constructJrd.ts
```diff
@@ -19,6 +19,10 @@ export default async function constructJrd(
   next: NextFunction,
 ): Promise<void> {
   const payId = req.query.resource
+
+  // Query parameters could be a string or a ParsedQs, or an array of either.
```
hbergren, line 23: What actually is a `ParsedQ` type?

nkramer44: `interface ParsedQs { [key: string]: string | string[] | ParsedQs | ParsedQs[] }`
That TS notation always confuses me, but I think it's just a type safe wrapper around key value pairs?

PR #607, opened 2020-07-22
Refactor Admin API

hbergren:

Just some follow-on refactors after Florent's PATCH PR.

We:
- Remove the custom `.options()` routes
- Explicitly define our user routes as `/:payId` rather than `/*`
- Correctly use Express `.use()` functions to check request headers and create response headers that need to be on multiple request endpoints, rather than duplicating the middleware functions inside the routes.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

npm run test is still green.

### DIFF
--- a/src/middlewares/adminApiHeaders.ts
+++ b/src/middlewares/adminApiHeaders.ts
@@ -14,7 +14,7 @@ import { ParseError, ParseErrorType, ContentTypeError } from '../utils/errors'
  *
  * @throws A ParseError if the PayID-API-Version header is missing, malformed, or unsupported.
  */
-export function checkAdminApiVersionHeaders(
+export function checkRequestAdminApiVersionHeaders(
   req: Request,
   res: Response,
   next: NextFunction,
@@ -65,7 +65,7 @@ export function checkAdminApiVersionHeaders(
  * @param next - An Express next() function.
  * @throws A ParseError if the Content-Type header is missing, malformed, or unsupported.
  */
-export function checkContentType(
+export function checkRequestContentType(
   req: Request,
   _res: Response,
   next: NextFunction,
@@ -104,7 +104,7 @@ export function checkContentType(
  * @param res - An Express Response object.
  * @param next - An Express next() function.
  */
-export function addAcceptPatchHeader(
+export function addAcceptPatchResponseHeader(
   _req: Request,
   res: Response,
   next: NextFunction,
@@ -115,30 +115,6 @@ export function addAcceptPatchHeader(
    * The Accept-Patch response HTTP header advertises which media-type the server is able to understand for a PATCH request.
    */
   res.header('Accept-Patch', 'application/merge-patch+json')
-
-  next()
-}
-
-/**
- * A middleware adding an Allow header in the response.
- *
- * @param _req - An Express Request object. Unused parameters can be conventionally underscored to avoid the check.
- * @param res - An Express Response object.
- * @param next - An Express next() function.
- */
-export function addAllowHeader(
-  _req: Request,
-  res: Response,
-  next: NextFunction,
-): void {
-  /**
-   * Add this header to all responses. Advertising Support in OPTIONS.
-   * A server can advertise its support for the PATCH method by adding it to the listing of allowed methods in the
-   * "Allow" OPTIONS response header defined in HTTP/1.1.  The PATCH method MAY appear in the "Allow" header
-   * even if the Accept-Patch header is absent, in which case the list of allowed patch documents is not advertised.
-   * Https://tools.ietf.org/html/rfc5789#section-3.
-   */
-  res.header('Allow', 'GET, PUT, OPTIONS, DELETE, PATCH')
 
   next()
 }
--- a/src/middlewares/users.ts
+++ b/src/middlewares/users.ts
@@ -33,7 +33,7 @@ export async function getUser(
   next: NextFunction,
 ): Promise<void> {
   // TODO:(dino) Validate PayID
-  const payId = req.params[0].toLowerCase()
+  const payId = req.params.payId.toLowerCase()
 
   // TODO:(hbergren) More validation? Assert that the PayID is `https://` and of a certain form?
   // Could use a similar regex to the one used by the database.
@@ -121,7 +121,7 @@ export async function putUser(
 ): Promise<void> {
   // TODO:(hbergren) Validate req.body and throw a 400 Bad Request when appropriate
   // TODO(hbergren): pull this PayID / HttpError out into middleware?
-  const rawPayId = req.params[0]
+  const rawPayId = req.params.payId
   const rawNewPayId = req.body?.payId
   const addresses = req.body?.addresses
 
@@ -203,7 +203,7 @@ export async function deleteUser(
   next: NextFunction,
 ): Promise<void> {
   // TODO:(hbergren) This absolutely needs to live in middleware
-  const payId = req.params[0].toLowerCase()
+  const payId = req.params.payId.toLowerCase()
 
   // TODO:(hbergren) More validation? Assert that the PayID is `https://` and of a certain form?
   // Do that using a regex route param in Express? Could use a similar regex to the one used by the database.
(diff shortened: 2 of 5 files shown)


src/middlewares/adminApiHeaders.ts
```diff
@@ -14,7 +14,7 @@ import { ParseError, ParseErrorType, ContentTypeError } from '../utils/errors'
  *
  * @throws A ParseError if the PayID-API-Version header is missing, malformed, or unsupported.
  */
-export function checkAdminApiVersionHeaders(
+export function checkRequestAdminApiVersionHeaders(
```
hbergren, line 17: rename function to be super explicit about what is going on.

src/middlewares/adminApiHeaders.ts
```diff
@@ -118,27 +118,3 @@ export function addAcceptPatchHeader(
...
 }
-
-/**
- * A middleware adding an Allow header in the response.
- *
- * @param _req - An Express Request object. Unused parameters can be conventionally underscored to avoid the check.
- * @param res - An Express Response object.
- * @param next - An Express next() function.
- */
-export function addAllowHeader(
```
hbergren, line 129: Refactored so the `Allow` header is auto-generated by Express once again.

src/middlewares/users.ts
```diff
@@ -33,7 +33,7 @@ export async function getUser(
   next: NextFunction,
 ): Promise<void> {
   // TODO:(dino) Validate PayID
-  const payId = req.params[0].toLowerCase()
+  const payId = req.params.payId.toLowerCase()
```
hbergren, line 36: This is a follow-up from changing the routes from `/*` to `/:payId`, to be explicit about what we expected here.

This is probably a pre-req to further Admin API work, like adding a `/:payId/addresses/:paymentNetwork/:environment` endpoint for CRUD operations, but seemed reasonable to fit into this refactor PR, since the new `PATCH` endpoint defined itself as `/:payId`.

src/routes/adminApiRouter.ts
```diff
@@ -22,61 +21,41 @@ const adminApiRouter = express.Router()
...
  */
 adminApiRouter
-  // Options user PayID route
-  .options(
-    '/:payId',
-    addAllowHeader,
-    addAcceptPatchHeader,
-    checkAdminApiVersionHeaders,
-    sendSuccess,
-  )
```
hbergren, line 32: Totally removed this `.options()` thing. We _needed_ a custom `.options()` endpoint so that it could respond with an `Accept-Patch` header, but I figured out that we could just do that in` .use('/:payId')`.

src/routes/adminApiRouter.ts
```diff
@@ -22,61 +21,41 @@ const adminApiRouter = express.Router()
...
+  // All /:payId requests should have an Accept-Patch response header with the PATCH mime type
+  .use('/:payId', addAcceptPatchResponseHeader)
```
hbergren, line 25: This adds the `Accept-Patch` header to all `/:payId` responses, as defined by the PATCH spec.

src/routes/adminApiRouter.ts
```diff
@@ -22,61 +21,41 @@ const adminApiRouter = express.Router()
...
+
+  // All [POST, PUT, PATCH] requests should have an appropriate Content-Type header,
+  // AND all Admin API requests should have an appropriate PayID-API-Version header.
+  .use('/*', checkRequestContentType, checkRequestAdminApiVersionHeaders)
```
hbergren, line 29: Now we no longer need `checkRequestContentType` and `checkRequestAdminApiVersionHeaders` in every single route middleware.

src/routes/adminApiRouter.ts
```diff
@@ -22,61 +21,41 @@ const adminApiRouter = express.Router()
...
-  )
+  .get('/:payId', wrapAsync(getUser), sendSuccess)
 
   // Create user route
   .post(
     '/',
     express.json(),
-    checkAdminApiVersionHeaders,
-    // TODO => checkContentType + E2E tests,
+    // TODO:(hbergren) checkRequestContentType E2E tests,
```
hbergren, line 38: Since we call `.use('/*', checkRequestContentType`), we are actually checking the content type here now, but we have not yet written E2E tests for that. That seemed more like it's own PR than part of this refactor.

src/routes/adminApiRouter.ts
```diff
@@ -22,61 +21,41 @@ const adminApiRouter = express.Router()
...
 
   // Get user route
-  .get(
-    '/*',
-    addAcceptPatchHeader,
-    checkAdminApiVersionHeaders,
-    wrapAsync(getUser),
-    sendSuccess,
-  )
+  .get('/:payId', wrapAsync(getUser), sendSuccess)
```
hbergren, line 32: Notice how we switched from `.get('/*')` to `.get('/:payId')`. That is why we switched from `req.params[0]` to `req.params.payId` in the `users.ts` middleware file.

test/integration/e2e/admin-api/optionsUsersPayId.test.ts
```diff
@@ -16,19 +16,28 @@ describe('E2E - adminApiRouter - OPTIONS /users/:payId', function (): void {
     app = await appSetup()
   })
 
-  it('Returns a 200 with the Allow header', function (done): void {
+  it('Returns a 200 with PATCH in the Allow header', function (done): void {
```
hbergren, line 19: Updated the test, because this is all we really care about.

test/integration/e2e/admin-api/optionsUsersPayId.test.ts
```diff
@@ -16,19 +16,28 @@ describe('E2E - adminApiRouter - OPTIONS /users/:payId', function (): void {
...
       .set('PayID-API-Version', payIdApiVersion)
-      .expect('Content-Type', /text\/plain/u)
       // THEN we expect the Allow header in the response
-      .expect('Allow', allowHeader)
+      .expect('Allow', /PATCH/u)
+      // AND we expect back a 200-OK
+      .expect(HttpStatus.OK, done)
+  })
+
+  it('Returns a 200 with the appropriate Accept-Patch response header', function (done): void {
```
hbergren, line 33: Added a new test to assert that OPTIONS appropriately responds with an `Accept-Patch`. (Originally this assertion was in the test above, but I split them).

src/routes/adminApiRouter.ts
austin-king: nice & clean

src/routes/adminApiRouter.ts
austin-king: cool, good comment

test/integration/e2e/admin-api/optionsUsersPayId.test.ts
dino-rodriguez: I thought we were dropping the options endpoint ?

src/routes/adminApiRouter.ts
hbergren: Yes. `/users/:payId` is an endpoint that allows `PATCH`ing, so all requests to that endpoint should advertise an `Accept-Patch` header so clients know what MimeType to use if they wanted to `PATCH /users/:payId`.

test/integration/e2e/admin-api/optionsUsersPayId.test.ts
hbergren: We did, but these are tests around `OPTIONS`.

The OPTIONS response is normally auto-generated by Express, but Florent made it explicit because we were having trouble getting OPTIONS to advertise the `Accept-Patch` header as required by the PATCH RFC.

However, by using `.use(/users/:payId, addAcceptPatchResponseHeader)`, we're able to get an OPTIONS response to advertise the header correctly. So these E2E tests are still worth having.

dino-rodriguez: Got it, that makes sense. Thanks for the explanation

PR #608, opened 2020-07-22
wrapAsync constructJrd call so error comes back correctly and bad req…

nkramer44:
* Improve error message in `constructJrd` middleware
* Made `constructJrd()` not async

### Context of Change
`constructJrd` used to be an async function because originally I was parsing the `discoveryLinks.json` file using `fs.promises.readFile`, which is an async function.  Now that we are importing the file as a ts module, this function doesn't need to be async.

### Type of Change

## Before / After
`constructJrd` was async, now it is not

## Test Plan
E2E testing for this endpoint coming in following PR.

### DIFF
--- a/src/routes/publicApiRouter.ts
+++ b/src/routes/publicApiRouter.ts
@@ -41,7 +41,7 @@ publicApiRouter
   .get('/status/health', sendSuccess)
 
   // PayID Discovery route
-  .get('/.well-known/webfinger', constructJrd, sendSuccess)
+  .get('/.well-known/webfinger', wrapAsync(constructJrd), sendSuccess)
 
   // Base PayID route
   .get(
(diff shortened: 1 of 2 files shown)


src/routes/publicApiRouter.ts
```diff
@@ -41,7 +41,7 @@ publicApiRouter
   .get('/status/health', sendSuccess)
 
   // PayID Discovery route
-  .get('/.well-known/webfinger', constructJrd, sendSuccess)
+  .get('/.well-known/webfinger', wrapAsync(constructJrd), sendSuccess)
```
hbergren, line 44: Why is `constructJrd` an async function?

(Not to say that `wrapAsync()` is wrong here, but it doesn't seem like `constructJrd` needs to be async).

nkramer44: Good question.  I don't think it needs to be.  It originally was because I was parsing the file with `fs.promises.parseFile` which is async, but now that Im just importing the file as a module I think it can be non-async

hbergren: Does it make sense to update `constructJrd` to be sync as well in this PR?

PR #609, opened 2020-07-22
Add E2E Tests for Discovery Endpoint

nkramer44:
* Adds two E2E integration tests to test PayID Discovery endpoint

### Context of Change
The `.well-known/webfinger` endpoint was previously untested.  Since the middleware is so lightweight, it doesn't really make sense to break things out into a service to unit test, and because middleware is so difficult to test, these e2e tests should suffice.

### Type of Change

## Before / After
No test coverage for Discovery -> integration test coverage for Discovery

## Test Plan
N/A

### DIFF
--- a/test/integration/e2e/public-api/discovery.test.ts
+++ b/test/integration/e2e/public-api/discovery.test.ts
@@ -1,0 +1,32 @@
+import HttpStatus from '@xpring-eng/http-status'
+import * as request from 'supertest'
+
+import App from '../../../../src/app'
+import { appSetup } from '../../../helpers/helpers'
+
+// eslint-disable-next-line node/file-extension-in-import -- typescript needs .json extension
+import * as discoveryLinks from './testDiscoveryLinks.json'
+
+let app: App
+
+describe('E2E - publicAPIRouter - PayID Discovery', function (): void {
+  // Boot up Express application and initialize DB connection pool
+  before(async function () {
+    app = await appSetup()
+  })
+
+  it('Discovery query returns JRD', function (done): void {
+    // GIVEN a PayID
+    const payId = 'alice$wallet.com'
+    const expectedResponse = {
+      subject: payId,
+      discoveryLinks,
+    }
+
+    // WHEN we make a GET request to the PayID Discovery endpoint
+    request(app.publicApiExpress)
+      .get(`/.well-known/webfinger?resource=${payId}`)
+      // THEN we get back a JRD with subject = the PayID and all links from the discoveryLinks.json file
+      .expect(HttpStatus.OK, expectedResponse, done)
+  })
+})
--- a/test/integration/e2e/public-api/discovery.test.ts
+++ b/test/integration/e2e/public-api/discovery.test.ts
@@ -16,10 +16,6 @@ describe('E2E - publicAPIRouter - PayID Discovery', function (): void {
     app = await appSetup()
   })
 
-  after(function () {
-    appCleanup(app)
-  })
-
   it('Discovery query returns JRD', function (done): void {
     // GIVEN a PayID
     const payId = 'alice$wallet.com'
@@ -48,5 +44,9 @@ describe('E2E - publicAPIRouter - PayID Discovery', function (): void {
       .get(discoveryPath)
       // THEN we get back a 400 with the expected error message
       .expect(HttpStatus.BadRequest, expectedErrorResponse, done)
+  })
+
+  after(function () {
+    appCleanup(app)
   })
 })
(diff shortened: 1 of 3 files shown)


test/integration/e2e/public-api/discovery.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    const payId = 'alice$wallet.com'
+    const expectedResponse = {
+      subject: payId,
+      discoveryLinks,
+    }
+
+    // WHEN we make a GET request to the PayID Discovery endpoint
+    request(app.publicApiExpress)
+      .get(`${discoveryPath}?resource=${payId}`)
+      // THEN we get back a JRD with subject = the PayID and all links from the discoveryLinks.json file
```
dino-rodriguez, line 35: Helpful test comments !

test/integration/e2e/public-api/discovery.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import * as request from 'supertest'
+
+import App from '../../../../src/app'
+import { appCleanup, appSetup } from '../../../helpers/helpers'
+
+// eslint-disable-next-line node/file-extension-in-import -- typescript needs .json extension
+import * as discoveryLinks from './testDiscoveryLinks.json'
+
+let app: App
+let discoveryPath: string
```
hbergren, line 11: Why not `const discoveryPath = '/.well-known/webfinger`?

test/integration/e2e/public-api/discovery.test.ts
```diff
@@ -0,0 +1,53 @@
...
+describe('E2E - publicAPIRouter - PayID Discovery', function (): void {
+  // Boot up Express application and initialize DB connection pool
+  before(async function () {
+    app = await appSetup()
+    discoveryPath = '/.well-known/webfinger'
+  })
+
+  after(function () {
+    appCleanup(app)
+  })
```
hbergren, line 22: Typically, we put the `after` hook at the end so the file reads better.

"Before the tests, run this... run the tests... after all tests are done, clean up."

test/integration/e2e/public-api/discovery.test.ts
```diff
@@ -0,0 +1,53 @@
+import HttpStatus from '@xpring-eng/http-status'
+import * as request from 'supertest'
+
+import App from '../../../../src/app'
+import { appCleanup, appSetup } from '../../../helpers/helpers'
+
+// eslint-disable-next-line node/file-extension-in-import -- typescript needs .json extension
+import * as discoveryLinks from './testDiscoveryLinks.json'
```
hbergren, line 8: Is the reason to use `./testDiscoveryLinks.json` that we would want a test to fail if we changed the `discoveryLinks.json` file, because the public interface of the app has changed?

test/integration/e2e/public-api/testDiscoveryLinks.json
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "rel": "https://payid.org/ns/payid-easy-checkout/1.0",
+    "template": "https://dev.wallet.xpring.io/wallet/xrp/testnet/payto?receiverPayId={receiverPayId}&amount={amount}&nextUrl={nextUrl}?easyCheckoutSuccess=true"
```
hbergren, line 4: Out of curiousity, why is the link to the `dev.wallet`? (I know it is in `discoveryLinks.json` as well).

test/integration/e2e/public-api/discovery.test.ts
nkramer44: へ‿(ツ)‿ㄏ I'll change

test/integration/e2e/public-api/discovery.test.ts
nkramer44: I usually like to separate test resources from src resources as much as possible to give more flexibility in testing.  `testDiscoveryLinks.json` and `discoveryLinks.json` happen to be the same right now, but you could imagine wanting to test something from the links file that you don't want to put in the actual links file.  It's probably unnecessary at this point, but I find making this distinction improves the reliability of tests.  

Feel free to push back on that

hbergren: That seems reasonable!

test/integration/e2e/public-api/testDiscoveryLinks.json
nkramer44: The `discoveryLinks.json` files we have committed to the public repo are kind of just placeholders/examples so we can spin up the server and do discovery locally.  In a real deployment, you'd want to mount an env specific file in the docker image or something.  I figured since these files would mostly be used for local testing/development, I might as well link to the dev wallet

PR #612, opened 2020-07-23
release script fixes

dino-rodriguez:

Fix dev environment not working on Docker for windows.

Fix minor error in bump script.

### Context of Change

Running the `devEnvUp` script on windows was not working with the env var.

We took on a dev dependency that allows cross-OS compatibility for scripts.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Add new dep, update scripts

## Test Plan

Can you test this on your windows box Florent?


package.json
```diff
@@ -20,19 +20,19 @@
...
+    "lint": "eslint . --ext .ts --fix --max-warnings 0 && prettier --write '**/*.{md,json}'",
+    "lintNoFix": "eslint . --ext .ts --max-warnings 0 && prettier --check '**/*.{md,json}'",
     "lintWatch": "chokidar src/**/*.ts -c \"npm run lint\" --initial --verbose",
     "prestart": "npm run build",
     "start": "node build/src/index.js",
     "startWatch": "tsc-watch --onSuccess 'npm run start'",
     "test": "nyc mocha 'test/**/*.test.ts'",
     "testWatch": "nyc mocha --watch 'test/**/*.test.ts'",
-    "devEnvUp": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose up -d",
+    "devEnvUp:default": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose up -d",
```
hbergren, line 31: Do you really need to specify `:default`? What if you didn't specify anything?

package.json
```diff
@@ -20,19 +20,19 @@
...
+    "devEnvUp:win32": "set PAYID_DOCKER_IMAGE_VERSION=%npm_package_version%&&docker-compose up -d",
```
hbergren, line 32: Have you tested this? 😄

dino-rodriguez:
@FlorentSG Could you test this out on windows to confirm it works?

package.json
dino-rodriguez: I tested the default command, and it worked, and just left a comment asking @FlorentSG if he could test windows.

I won't merge until I get a green light on windows.

package.json
austin-king: I'm a fan of not specifying anything for default as well

package.json
dino-rodriguez: Haha, I just realized I actually ran `npm run devEnvUp:default` 

The point is to just run `npm run devEnvUp` , and it can figure out which script to run

I updated the code to actually use `run-script-os`

dino-rodriguez:
@hbergren @austin-king So the way this works now:

you run `npm run devEnvUp`

if it's on windows, it uses the `windows` script

it it's on anything else, it uses the `default` script

package.json
dino-rodriguez: I need to specify `default`, I just wasn't even using `run-script-os` in this haha

See my comment below

hbergren: Should this technically be `devEnvUp:nix`?

dino-rodriguez: They both accomplish the same thing :

`nix` says run on `nix`

`default` run on anything that isn't explicitly declared

The split is `windows` and everything else

package.json
FlorentSG: Hi @dino-rodriguez , I just tested it on Windows 10 and it works. I simply ran `npm run devEnvUp`.

package.json
```diff
@@ -28,7 +28,9 @@
     "startWatch": "tsc-watch --onSuccess 'npm run start'",
     "test": "nyc mocha 'test/**/*.test.ts'",
     "testWatch": "nyc mocha --watch 'test/**/*.test.ts'",
-    "devEnvUp": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose up -d",
+    "devEnvUp": "run-script-os",
+    "devEnvUp:default": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose up -d",
+    "devEnvUp:win32": "set PAYID_DOCKER_IMAGE_VERSION=%npm_package_version%&&docker-compose up -d",
     "devDbUp": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose up -d db",
```
FlorentSG, line 34: Hi @dino-rodriguez , this will create the same error message on Windows as it is a Linux command.

We can do two things:
1. Remove `PAYID_DOCKER_IMAGE_VERSION=$npm_package_version` from the command as this variable is not needed to create the database only. By doing that, there will be a `warning` displayed on both Mac and Windows but the database container will start fine.


2. If we don't want to see this warning then we need to create a specific script for Linux and Windows as we did for `npm run devEnvUp`.
So we would have:
```
"devDbUp": "run-script-os",
"devDbUp:default": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose up -d db",
"devDbUp:win32": "set PAYID_DOCKER_IMAGE_VERSION=%npm_package_version%&&docker-compose up -d db",
```

I personally prefer the option 1 as it keeps the number of scripts "low" and the warning doesn't affect anything.

FlorentSG, line 34: Hi @dino-rodriguez ,`npm run devDbUp` will create the same error message on Windows as it is a Linux command.

We can do two things:

1. Remove PAYID_DOCKER_IMAGE_VERSION=$npm_package_version from the command as this variable is not needed to create the database only. By doing that, there will be a warning displayed on both Mac and Windows but the database container will start fine.

2. If we don't want to see this warning then we need to create a specific script for Linux and Windows as we did for npm run devEnvUp.
So we would have:

"devDbUp": "run-script-os",
"devDbUp:default": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose up -d db",
"devDbUp:win32": "set PAYID_DOCKER_IMAGE_VERSION=%npm_package_version%&&docker-compose up -d db",

I personally prefer option 1 as it keeps the number of scripts "low" and the warning doesn't affect anything.

package.json
```diff
@@ -28,7 +28,9 @@
...
     "devDown": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose down",
```
FlorentSG, line 35: Hi @dino-rodriguez , `npm run devDown` will also trigger the same error message on Windows as it is a Linux command.

We can do the same two things as for `npm run devDbUp`:

1. Remove `PAYID_DOCKER_IMAGE_VERSION=$npm_package_version` from the command as this variable is not needed to remove the containers and network. By doing that, there will be a `warning` displayed on both Mac and Windows but the containers will be deleted.

warning


2. If we don't want to see this warning then we need to create a specific script for Linux and Windows as we did for `npm run devEnvUp`.
So we would have:

```
"devDown": "run-script-os",
"devDown:default": "PAYID_DOCKER_IMAGE_VERSION=$npm_package_version docker-compose down",
"devDown:win32": "set PAYID_DOCKER_IMAGE_VERSION=%npm_package_version%&&docker-compose down",
```

I personally prefer option 1 as it keeps the number of scripts "low" and the warning doesn't affect anything.

FlorentSG:
Hi @dino-rodriguez , I couldn't add a comment for the `build` script but this one also doesn't work on Windows due to the `rm` command which is Linux specific. As the `prestart` script uses `build` , an error is thrown when a user runs `npm start`.

We need to create a specific `build` script for Windows as well:

```
"build": "run-script-os",
"build:default": "rm -rf build && tsc --project . && npm run copySQL && npm run copyHTML && npm run copyJSON",
"build:win32": "del /q build && tsc --project . && npm run copySQL && npm run copyHTML && npm run copyJSON",
```

/q = Quiet mode; suppresses prompts for delete confirmations.

https://www.lifewire.com/delete-command-2625859

package.json
dino-rodriguez: I prefer this too -- I'm going with this. Thanks for the suggestion.

package.json
dino-rodriguez: I prefer option #2 as well. I'm going to remove the env var from both of these.

dino-rodriguez:
@FlorentSG Thanks! I've incorporated the build script suggestions in there as well.

FlorentSG:
You're welcome @austin-king 
@dino-rodriguez , I tested again one last time on Windows and `npm run devEnvUp`, `npm run devDbUp`, `npm run devDown`, `npm start` work fine.

I don't know how to use the `bump` and `release` script but I think they may generate an issue on Windows.
Nevertheless, I think only you @dino-rodriguez , @austin-king and @hbergren are using them at the moment so I am going to approve this PR. It's just something to think about if someone needs to use them on Windows in the future.

PR #618, opened 2020-07-27
refactor: metrics library

dino-rodriguez:

Swaps out the core metrics service for @payid-org/payid-metrics.

### Context of Change

We wanted to pull metrics into a library so others can integrate it into their
PayID servers.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Still the same code, so the same tests. I removed the unit tests, and put those
in the metrics library.


test/helpers/helpers.ts
```diff
@@ -12,7 +12,7 @@ import syncDatabaseSchema from '../../src/db/syncDatabaseSchema'
  *
  * @returns A deeply cloned object.
  */
-export function structuredClone<T>(obj: T): T {
```
dino-rodriguez, line 15: Only used as helper internally now ( used in `checkMetricsConfiguration` )

dino-rodriguez:
@hbergren Ok, this should now use metrics `1.0`.

src/config.ts
```diff
@@ -47,7 +47,7 @@ const config = {
      *
      * This will be dynamically set by incoming requests if the ENV var is unset.
      */
-    domain: process.env.PAYID_DOMAIN,
+    domain: process.env.PAYID_DOMAIN ?? 'missing_domain',
```
hbergren, line 50: So is `PAYID_DOMAIN` a required env var again? Did you cut auto-filling this from the first Public API request?

dino-rodriguez: No, we still autofill, but now we check for `missing_domain` : https://github.com/payid-org/payid/pull/618/files#diff-5d5cd96e1b9c48577239a06f8cf1a202R24

hbergren: Do you you feel that is easier to understand than what was there before?

I guess I wrote the previous code which is why I grok it haha.

dino-rodriguez:
@hbergren I did this because `config.domain` is now required, as opposed to `undefined | string`. This was brought up here : https://github.com/payid-org/server-metrics/pull/1#discussion_r461056853

src/config.ts
dino-rodriguez: I did this because `config.domain` is now required, as opposed to `undefined | string`. So metrics wouldn't accept the prior configuration. This was brought up here : https://github.com/payid-org/server-metrics/pull/1#discussion_r461056853

hbergren: Got it... Oh TypeScript. I love you less each day.

PR #619, opened 2020-07-27
extend account table with identity_key and address table with identity_key_signature [1/x]

austin-king:

Completed in both schema definitions and appended migration files:
- Add `identity_key` to `account` table
- Add `identity_key_signature` to `address` table
- Extend TS database types to reflect new DB schemas

Asana task: https://app.asana.com/0/1145813803187951/1186312611097768

I set the varchar length as 600 for both because we are not entirely certain what signing algorithms people may use for this feature (we know what we will use, but not everybody else) so I opted for a pretty liberal varchar length.

### Context of Change

Part of a series of PRs to implement user supplied signatures verifying ownership of a blockchain address.
For the eng. spec, see: https://docs.google.com/document/d/1swKkzY0zTl1JuqfH6if1qbdo3l-y9up97XxDTX-80WA/edit#
For the high level asana task, see: https://app.asana.com/0/1145813803187951/1186312611097763

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Database tables are now ready to store user supplied signatures 

## Test Plan

No tests added in this PR, as I continue implementing this high level feature I will write code that depends on these database tables (& corresponding unit tests), ensuring that they work as expected.

## Future Tasks

- Expand the current Public API GET response
- Expand the PayID Admin API to accept signed addresses


src/db/schema/01_account.sql
```diff
@@ -4,6 +4,7 @@
 CREATE TABLE IF NOT EXISTS account (
 	id uuid PRIMARY KEY DEFAULT(gen_random_uuid()),
 	pay_id varchar(200) UNIQUE NOT NULL,
+        identity_key varchar(600),
```
hbergren, line 7: nit: replace the spaces with a single tab.

Should `identity_key` be unique?

src/db/migrations/07_add_identity_key.sql
```diff
@@ -0,0 +1,7 @@
+-- Add identity key column to account
+--
+-- This column is used to store user generated identity keys so that clients
+-- querying addresses can verify the individual user confirmed ownership
+-- of an address via a signature.
+ALTER TABLE account
+ADD COLUMN IF NOT EXISTS identity_key varchar(600);
```
hbergren, line 7: Maybe we should even remove the limit from the `varchar` columns.

They won't actually take up any more room in the database (in Postgres at least). So this limit is really just a "semantic" check, but we actually don't know the reasonable maximum for `identity_key` and `signature`.

src/db/schema/01_account.sql
austin-king: I originally had it as `unique` but then removed it for the edge case in which you have 1 human using multiple accounts on the same "wallet service" and they use the same identity key. That is edge case enough that I'd be down to make it `unique` if you think that would protect other mishaps though.

hbergren: Ok, that's a reasonable argument for why this shouldn't be unique.

src/db/migrations/07_add_identity_key.sql
austin-king: Think we should do this just for the new `identity_key` and `identity_key_signature` columns or _all_ varchar length limits?

src/db/schema/01_account.sql
austin-king: Got it, also fixed tabs issue

src/db/migrations/07_add_identity_key.sql
hbergren: Eh. Maybe we should remove limits for everything, but the limits for `payid` and address stuff is pretty reasonable. The thing I'm hesitant about with keys and signatures is that I really have no idea what a reasonable upper limit is on a base 64 encoded key or signature is.

src/db/schema/02_address.sql
```diff
@@ -7,6 +7,8 @@ CREATE TABLE IF NOT EXISTS address (
 
 	details jsonb NOT NULL,
 
+        identity_key_signature varchar(600),
```
hbergren: nit: tab.

src/db/migrations/07_add_identity_key.sql
austin-king: Ok, sound good, I'll just remove limits for this PR and not mutate the old fields then.

austin-king: pushed

PR #621, opened 2020-07-28
add verified address interface [2/x]

austin-king:

- adds VerifiedAddress type as defined in this [gist](https://gist.github.com/austin-king/aa7f830899354b283f93527a8d270782) that we agreed upon in the [design doc](https://docs.google.com/document/d/1swKkzY0zTl1JuqfH6if1qbdo3l-y9up97XxDTX-80WA/edit#).

### Context of Change

A baby step on our way to implementing non-custodial verifiable PayID. I'm adding this interface before I begin work on the Admin API.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

- 1 new file `src/types/verifiedAddress.ts` containing `VerifiedAddress` interface
- adds export of `Address` interface from `src/db/publicAPI.ts` to be used in `VerifiedAddress` interface

## Test Plan

No tests added in this PR, will be tested as I begin implementing logic in subsequent PRs

## Future Tasks

My subsequent PRs implementing logic for the Admin API will use this interface to ensure that `POSTs` conform to our expected interfaces.


src/db/migrations/07_add_identity_key.sql
```diff
@@ -0,0 +1,7 @@
+-- Add identity key column to account
```
austin-king: Ignore all files in this PR except the last 2: `src/types/publicAPI.ts` & `src/types/verifiedAddress.ts`.

Once my [previous PR](https://github.com/payid-org/payid/pull/619) is merged I will update this PR from master and the only diff will be those remaining 2 files

src/types/verifiedAddress.ts
```diff
@@ -0,0 +1,36 @@
+import { Address } from './publicAPI'
+
+/**
+ * Object containing address information alongside signatures.
+ */
+// eslint-disable-next-line import/no-unused-modules -- will disable in next PR
```
austin-king, line 6: Will delete this disable comment when I implement the subsequent PRs containing the logic that depends on this interface

hbergren, line 6: Alternatively, you could remove the `export` statement, and just add the `export` in the next PR.

Either way.

src/types/verifiedAddress.ts
```diff
@@ -0,0 +1,36 @@
...
+export interface VerifiedAddress {
+  readonly payload: VerifiedAddressPayload
+  readonly signatures: VerifiedAddressSignature[]
```
hbergren, line 9: I think we would want to make the array readonly as well, right?

```suggestion
  readonly signatures: readonly VerifiedAddressSignature[]
```

src/types/verifiedAddress.ts
```diff
@@ -0,0 +1,36 @@
...
+interface VerifiedAddressPayload {
+  payId: string
+  payIdAddress: Address
+}
+
+/**
+ * Object containing a signature corresponding to an address.
+ *
+ * Protected -- base64 encoded identity key
+ * signature -- base64 encoded signatureA.
```
hbergren, line 21:
```suggestion
 * signature -- base64 encoded signature.
```

src/types/verifiedAddress.ts
```diff
@@ -0,0 +1,36 @@
...
+
+interface VerifiedAddressPayload {
+  payId: string
+  payIdAddress: Address
+}
+
+/**
+ * Object containing a signature corresponding to an address.
+ *
+ * Protected -- base64 encoded identity key
```
hbergren, line 20:
```suggestion
 * protected -- base64 encoded identity key.
```

src/types/verifiedAddress.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Object containing address information alongside signatures.
+ */
+// eslint-disable-next-line import/no-unused-modules -- will disable in next PR
+export interface VerifiedAddress {
+  readonly payload: VerifiedAddressPayload
+  readonly signatures: VerifiedAddressSignature[]
+}
+
+interface VerifiedAddressPayload {
```
dino-rodriguez, line 12: Is the reason we are not using `PaymentInformation` because it is an array of addresses?

src/types/verifiedAddress.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/**
+ * Object containing a signature corresponding to an address.
+ *
+ * Protected -- base64 encoded identity key
+ * signature -- base64 encoded signatureA.
+ */
+interface VerifiedAddressSignature {
+  name: VerifiedAddressSignatureType
```
hbergren, line 24: I think name is going to live inside the `protected` header.

David added an Asana task for me to do that, and I updated the [gist](https://gist.github.com/hbergren/8dd95826dddefc49d8635366031670dd) accordingly.

So you can probably remove `name` as a standalone field.

src/types/verifiedAddress.ts
```diff
@@ -0,0 +1,36 @@
...
+  protected: string
+  signature: string
+}
+
+/**
+ * Identifies the type of signature being used, as of now we are only
+ * implementing IdentityKey signatures, but future types like
+ * WebPKI or ProofOfControl should be added to this enum.
+ */
+enum VerifiedAddressSignatureType {
```
hbergren, line 34: I guess this enum is unused if we don't have the `name` property, but I think it may make sense to leave it as a form of documentation?

Also, when we have multiple signatures (Proof of Control?) even if we store the base64 encoded string in the DB, we may need to parse it to determine which key/signature pair to store in which column.

src/types/verifiedAddress.ts
austin-king: Ya, I did that then build failed because it was an unused var so I was getting an error either way

src/db/migrations/07_add_identity_key.sql
hbergren: I just updated the base branch for this PR to your other PR branch.

If you merge the first PR into `master`, this will auto-update, and it lets the diff be more readable.

src/types/verifiedAddress.ts
austin-king: No, we're not using it b/c it does not contain the required fields like `payIdAddress` (used to distinguish between JWT `address` field)

src/types/verifiedAddress.ts
austin-king: I'll document / add it back when I implement parsing logic to ensure a signature matches a type we have implemented. Irrelevant to this PR though now, so removing.

src/types/verifiedAddress.ts
austin-king: cool, removed

src/types/database.ts
```diff
@@ -25,6 +25,8 @@ export interface Address {
 
   readonly identityKeySignature?: string
 
+  readonly identityKeySignature?: string
```
hbergren, line 28: This looks like it got duplicated?

src/types/verifiedAddress.ts
```diff
@@ -0,0 +1,24 @@
...
+
+interface VerifiedAddressPayload {
+  payId: string
+  payIdAddress: Address
+}
+
+/**
+ * Object containing a base64 encoded identity key (protected) alongside a
+ * corresponding signature.
+ */
```
hbergren, line 20: Is it worth saying this is just using JWS?

src/types/database.ts
austin-king: Ya lol sorry was in the middle of merging from master & had to hop into a 1:1 w/ David so left this in a bad state

src/db/migrations/07_add_identity_key.sql
austin-king: Got it, will do in future PRs. Thanks for the tip!

src/types/verifiedAddress.ts
austin-king: Yup, adjusted comment

src/middlewares/constructJrd.ts
```diff
@@ -1,5 +1,6 @@
 import { NextFunction, Request, Response } from 'express'
 
+// eslint-disable-next-line node/file-extension-in-import -- necessary for JSON
```
austin-king, line 3: Somehow this import got merged into master even though it throws a linter error, so I disabled it. I tried removing the `.json` but then TS can't recognize and import it with the error:

`Cannot find module './testDiscoveryLinks' or its corresponding type declarations.`

Disabling for now, @hbergren do you have a suggestion for a better way to do this?

PR #623, opened 2020-07-29
feat: e2e testing content-type and accept-patch

florent-uzio:

Added E2E testing for Content-Type and Accept-Patch on the GET, POST, PUT and DELETE routes.

### Context of Change

We need to make sure the correct Headers are included (or not) in the different routes.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Added and updated the E2E tests for the GET, POST, PUT and DELETE routes.
Before the Content-Type and Accept-Patch header were not tested on these routes.

## Test Plan

I ran `npm run test` and made sure all of them passed.


test/integration/e2e/admin-api/postUsers.test.ts
```diff
@@ -43,8 +44,18 @@ describe('E2E - adminApiRouter - POST /users', function (): void {
...
-      .expect(HttpStatus.Created, done)
+      // THEN we expect back a 201 - CREATED
+      .expect(HttpStatus.Created)
+      .end(function (err, res) {
+        if (err) {
+          return done(err)
+        }
+
+        // AND ensure Accept-Patch header does not exist
+        expect(res.header).to.not.have.key('Accept-Patch')
```
hbergren, line 55: This `.end()` test is particularly verbose, so I would prefer if instead of adding this to every test case, we just created a single new test that verified this.

test/integration/e2e/admin-api/putUsers.test.ts
```diff
@@ -281,10 +299,91 @@ describe('E2E - adminApiRouter - PUT /users', function (): void {
...
+      .expect('Accept-Patch', acceptPatch)
       // THEN we expect back a 409 - CONFLICT and our expected error response
       .expect(HttpStatus.Conflict, expectedErrorResponse, done)
   })
 
+  it('Returns a 415 - Unsupported Media Type when sending a wrong Content-Type header', function (done): void {
+    // GIVEN a PayID known to resolve to an account on the PayID service
+    const payId = 'alice$xpring.money'
+    // We need to send the body as a String if Content-Type is not application/json.
+    // Otherwise an error ERR_INVALID_ARG_TYPE will be thrown by Supertest in the send() method.
```
hbergren, line 312: Really helpful comment. Thank you.

test/integration/e2e/admin-api/postUsers.test.ts
florent-uzio: Hi @hbergren , good point, I made the relevant modifications and added only 1 test that verifies the POST method doesn't add an `Accept-Patch` header in the response.

florent-uzio:
@hbergren , it's not really the goal of this PR but I knew I had to make a small change in a comment in `docker-compose.yml` and I took the opportunity to do it here.
Thanks

test/integration/e2e/admin-api/postUsers.test.ts
```diff
@@ -43,10 +44,54 @@ describe('E2E - adminApiRouter - POST /users', function (): void {
...
+      // THEN we expect the Location header to be set to the path of the created user resource
+      .expect('Location', `/users/${userInformation.payId}`)
+      // THEN we expect back a 201 - CREATED
+      .expect(HttpStatus.Created)
+      .end(function (err, res) {
+        if (err) {
+          return done(err)
+        }
+
+        // AND ensure Accept-Patch header does not exist
```
dino-rodriguez, line 88: Should this be `AND` or another `THEN`? Providing the suggestion just in case.

```suggestion
        // THEN ensure Accept-Patch header does not exist
```

hbergren:
Thanks again @florent-uzio !

PR #625, opened 2020-07-30
Fix Admin API Get

hbergren:

Fix a bug so that if you GET a user on the Admin API that has no associated addresses, you now get a 200 with a payload with an empty address array rather than a 404 - Not Found.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added a test specifically for this case.


src/data-access/users.ts
```diff
@@ -7,6 +7,27 @@ import knex from '../db/knex'
...
+ * Checks if a given PayID exists in the account table in the PayID database.
+ *
+ * @param payId - The PayID to insert in the account table.
+ *
+ * @returns A boolean indicating whether the PayID exists.
+ *
+ * Note: This could actually be done in getAllAddressInfoFromDatabase,
+ * if we changed the return type a bit, which would let us always do a single database call,
+ * instead of having to call this as a follow-up check, but it's probably cleaner to
+ * let them have separate concerns until performance becomes an issue.
```
hbergren, line 20: This is an important note. Basically, separation of concerns trumps performance (and I save myself a much bigger refactor).

src/db/seed/seeded_values_for_testing.sql
```diff
@@ -7,7 +7,8 @@ INSERT INTO account(id, pay_id) VALUES
 ('223ece9c-2a15-48e1-9df6-d9ac77c5db90', 'bob$127.0.0.1'),
 ('ec06236e-d134-4a7b-b69e-0606fb54b67b', 'alice$xpring.money'),
 ('69b0d20a-cdef-4bb9-adf9-2109979a12af', 'bob$xpring.money'),
-('b253bed2-79ce-45d0-bbdd-96867aa85fd5', 'zebra$xpring.money');
+('b253bed2-79ce-45d0-bbdd-96867aa85fd5', 'zebra$xpring.money'),
+('8a75f884-ab16-40c4-a82a-aca454dad6b2', 'empty$xpring.money');
```
hbergren, line 11: Added a new PayID for testing, with no associated addresses.

test/integration/data-access/reports.test.ts
```diff
@@ -46,7 +46,7 @@ describe('Data Access - getPayIdCounts()', function (): void {
 
   it('getPayIdCount - Returns a count of PayIDs', async function () {
     const payIdCount = await getPayIdCount()
-    const expectedPayIdCount = 5
+    const expectedPayIdCount = 6
```
hbergren, line 49: Added a PayID.

test/integration/e2e/admin-api/metrics.test.ts
```diff
@@ -100,7 +100,7 @@ describe('E2E - adminApiRouter - GET /metrics', function (): void {
 
     // We create 8 PayIDs in the tests before this one,
     // and start with 5 seeded PayIDs, for a total of 13.
-    await assertMetrics(/actual_payid_count\{org="127.0.0.1"\} 13/u)
+    await assertMetrics(/actual_payid_count\{org="127.0.0.1"\} 14/u)
```
hbergren, line 103: Added a PayID.

src/data-access/users.ts
dino-rodriguez: I agree, no need to over-index on performance over developer hours if performance is not an issue.

hbergren: Just to stick in your head, the way you would do this is a left/right join instead of an inner join in that function.

That way, having no addresses wouldn't filter out the PayID if it actually existed.

src/data-access/users.ts
```diff
@@ -7,6 +7,27 @@ import knex from '../db/knex'
...
+ */
+// TODO:(hbergren): Type payId better
+export async function checkUserExistence(payId: string): Promise<boolean> {
+  const result = await knex.select(
+    knex.raw('exists(select 1 from account where pay_id = ?)', payId),
+  )
+
+  return result[0].exists
```
austin-king, line 28: I didn't know `.exists` was a function, nice and simple

hbergren: It's not a function per se, (well, it is in SQL).

So above I do `knex.raw('exists(...)')`

And thus, `exists` is a column on the return table that holds a boolean, and I can just access that column as a property on the returned first row.

PR #627, opened 2020-07-30
Update Discovery to latest version of Easy Checkout

nkramer44:
* Conforms the JRD that we pass back for Discovery queries to the newest version of the Easy Checkout RFC.
* Fixes a bug in constructing the JRD where the "links" field was being sent as "discoveryLinks"

### Context of Change
Making [this](https://github.com/xpring-eng/chef-xrp/pull/16) change to ChefXRP and [this](https://github.com/xpring-eng/xpring-wallet-system/pull/913) change to xpring.io to update everything to the latest and greatest of Easy Checkout.

The JRD format no longer includes a "template", which has been replaced by "href".

The bug was just a product of me being a TS novice.  I was constructing the JRD like 
```typescript
{
 subject: payId,
 discoveryLinks
}
```
so the payload returned had "discoveryLinks" as the label for the array of links, when it should be "links"

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
Conformed to earlier iteration of PayID Easy Checkout -> Now up to date with current version

## Test Plan
Updated e2e integration tests.  Have tested the whole Easy Checkout flow with ChefXRP and the xpring wallet locally, and all seems to work.

## Future Tasks
Note that I am still waiting on [this PR](https://github.com/xpring-eng/payid-charts/pull/11) to configure k8s to mount the `discoveryLinks.json` file based on environment.  Until that gets through, we need to make the change here and all environments will have the same link href.


nkramer44:
@hbergren any idea why the Link Checker is failing? I see it got a 504 checking https://payid.org/.  Not sure if it was just bad timing, but I don't see anywhere to re-run CI

PR #635, opened 2020-08-03
feat: verifiedAddresses in public API interface

dino-rodriguez:

Add `verifiedAddress` field to `PaymentInformation`.

Update Public API interface to adhered to updated `PaymentInformation` type.

Update tests so they continue to pass with the expanded interface.

### Context of Change

The goal of this PR is to add `verifiedAddresses` to the type definition, and to
subsequently update all places in the code so they code compiles & tests
continue to pass.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Before: `PaymentInformation` does not contain `verifiedAddresses`
After: `PaymentInformation` contains `verifiedAddresses`

## Test Plan

This PR updates all tests so they continue to pass.

I will add any necessary new tests in a subsequent PR.

### DIFF
--- a/src/services/basePayId.ts
+++ b/src/services/basePayId.ts
@@ -29,6 +29,7 @@ export function formatPaymentInfo(
         addressDetails: address.details,
       }
     }),
+    verifiedAddresses: [],
     ...(payId && { payId }),
   }
 
--- a/src/types/publicAPI.ts
+++ b/src/types/publicAPI.ts
@@ -1,3 +1,8 @@
+/* eslint-disable import/no-cycle --
+Cycle between this file and types/verifiedAddress.ts. Should we combine these into one?
+*/
+import { VerifiedAddress } from './verifiedAddress'
+
 /**
  * Type of payment address in PaymentInformation.
  */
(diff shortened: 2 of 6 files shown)


src/services/basePayId.ts
```diff
@@ -29,6 +29,7 @@ export function formatPaymentInfo(
         addressDetails: address.details,
       }
     }),
+    verifiedAddresses: [],
```
dino-rodriguez, line 32: For now, this will just be an empty response.

src/types/publicAPI.ts
```diff
@@ -1,3 +1,8 @@
+/* eslint-disable import/no-cycle --
+Cycle between this file and types/verifiedAddress.ts. Should we combine these into one?
```
dino-rodriguez, line 2: Currently we have a types file for verified addresses, and one for the public API. 

The other two type files are database types.

I'm thinking that the verified addresses types, and public api types, are really all just protocol types.

I'm thinking we should combine these into one file like (`types/protocol.ts` or `types/address.ts`). This would require a bit of a refactor to all the references to previous types files, so I would do that as a subsequent PR.

What do you think?

test/integration/e2e/public-api/basePayId.test.ts
```diff
@@ -31,6 +31,7 @@ describe('E2E - publicAPIRouter - Base PayID', function (): void {
           },
         },
       ],
+      verifiedAddresses: [],
```
dino-rodriguez, line 34: Updates a bunch of tests so they continue to pass.

src/types/publicAPI.ts
austin-king: Ya I'm into that, these have been calling for a refactor for a minute now

PR #636, opened 2020-08-03
feat: AdminAPI GET for verified accounts

austin-king:

Admin API can now return `identity_key` field for verified accounts

### Context of Change

Series of PRs implementing verifiable PayID

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

AdminAPI now returns identity keys

## Test Plan

Previous tests continue to pass + 1 new one to confirm it returns as expected


src/db/seed/seeded_values_for_testing.sql
```diff
@@ -10,6 +10,9 @@ INSERT INTO account(id, pay_id) VALUES
 ('b253bed2-79ce-45d0-bbdd-96867aa85fd5', 'zebra$xpring.money'),
 ('8a75f884-ab16-40c4-a82a-aca454dad6b2', 'empty$xpring.money');
 
+INSERT INTO account(id, pay_id, identity_key) VALUES
+('27944333-faf6-41e8-90c3-1ec9001f0830', 'verified$example.com', 'aGkgbXkgbmFtZSBpcyBhdXN0aW4gYW5kIEkgYW0gdGVzdGluZyB0aGluZ3M=');
```
austin-king, line 14: Needed one of these for my test, feel free to build further tests off of this PayID or add more verified payIDs

src/data-access/payIds.ts
```diff
@@ -18,3 +18,20 @@ export default async function getAllAddressInfoFromDatabase(
...
   return addressInformation
 }
+
+/**
+ * Retrieves the identity key for a specific PayID.
+ *
+ * @param payId - The PayID that we are requesting an identityKey for.
+ * @returns The identity key for that PayID if it exists.
+ */
+export async function getIdentityKeyFromDatabase(
```
dino-rodriguez, line 28: Just so I am clear on what this is supposed to be, is this supposed to be the base64 encoded `protected` field from this gist: https://gist.github.com/hbergren/8dd95826dddefc49d8635366031670dd ?

src/db/seed/seeded_values_for_testing.sql
dino-rodriguez: Sweet. Is this a valid protected field? As in, would this pass if you ran `jws.verify` on it? 

I was thinking that this would be nice to have seeded values with legit payloads to test that the reference actually works as expected with JWS.

To be clear - don't think this needs to be in this PR. But just wondering, as I may consider updating this to be a valid payload as I am testing stuff in the public api.

src/middlewares/users.ts
```diff
@@ -55,12 +58,20 @@ export async function getUser(
   }
 
   const addresses = await getAllAddressInfoFromDatabase(payId)
-
-  res.locals.response = {
-    payId,
-    addresses,
+  const identityKey = await getIdentityKeyFromDatabase(payId)
+
+  if (identityKey === null) {
```
dino-rodriguez, line 63: Nice, no need to have null values returned here.

src/data-access/payIds.ts
austin-king: Yes I believe so. I know for certain it contains the jwk sub-object, but I also believe that those other fields are included.

austin-king:
@dino-rodriguez nope the base64 encoded values put in the DB are bogus right now. I agree, we'll likely want to swap them for real ones eventually. Let's punt on that for now to make sure we hit the sprint goals & incorporate that validation testing into our next sprint.

src/middlewares/users.ts
florent-uzio: @austin-king would it also make sense to check if `identityKey` has at least one character?

I created a PayID with `"identityKey": ""` and the response shows the `identityKey` without any value.
```
{
    "payId": "flo2$127.0.0.1",
    "identityKey": "",
    "addresses": [
        {
            "paymentNetwork": "XRPL",
            "environment": "TESTNET",
            "details": {
                "address": "TVnGpXXZZ3xAZfhT42ntuCR4Uh3Rv9LE4BcZJeH1zds2CQ2"
            }
        },
        {
            "paymentNetwork": "BTC",
            "environment": "MAINNET",
            "details": {
                "address": "1F1tAaz5x1HUXrCNLbtMDqcw6o5GNn4xqX"
            }
        }
    ]
}
```

Shall we write:
`if (identityKey === null || identityKey.length === 0) {`

austin-king: Good thinking @florent-uzio, just added that in alongside a test for it.

test/integration/e2e/admin-api/metrics.test.ts
```diff
@@ -100,7 +100,7 @@ describe('E2E - adminApiRouter - GET /metrics', function (): void {
 
     // We create 8 PayIDs in the tests before this one,
     // and start with 5 seeded PayIDs, for a total of 13.
```
florent-uzio, line 102: A small detail, but is the total of 13 still accurate? Or should it be 16?
I don't know if the others numbers are correct as well (8 PayIDs and 5 seeded PayIDs)

austin-king: Good point, going to update that comment and merge.

PR #639, opened 2020-08-04
feat: make fiat change non-breaking

dino-rodriguez:

Updates public API response to intelligently decide on `ACH` vs. `FIAT`
depending on whether the request is a `1.0` or `1.1` response.

Rolls forward the PayID protocol server version to `1.1`.

### Context of Change

Makes ACH -> FIAT `addressDetailsType` change non-breaking.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

We know do some backwards compatibility stuff.

## Test Plan

I fixed all existing tests, but need to add more test coverage now in a
subsequent PR.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,4 +1,4 @@
-export const payIdServerVersions: readonly string[] = ['1.0']
+export const payIdServerVersions: readonly string[] = ['1.0', '1.1']
 export const adminApiVersions: readonly string[] = ['2020-05-28']
 
 /**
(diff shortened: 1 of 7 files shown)


nkramer44:
v1.1.0 uses `AchAddressDetails` right? If so, I think it should be "if not v1.0.0 or v1.1.0, send FiatAddressDetails, otherwise send AchAddressDetails".  Could be wrong, but I think the way you have it squeezes the change into v1.1.0, making v1.1.0 backwards compatible with v1.0.0, not v1.2 backwards compatible with v1.1.0

PR #640, opened 2020-08-05
feat: verifiable payid public api-data-access integration

dino-rodriguez:

Adds data-access functions to PayID Public API.

Disambiguates between a missing PayID and missing addresses for PayIDs with
different error messages & updating the metrics logging.

Refactors some of error handling to make this "friendlier" for V. PayID integration.

### Context of Change

Thanks to the identity key data-acccess work, I was able to solve this bug. 

The commment in the code explains how.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Updates metrics, adds V.PayID data access, moves some metrics capture to error
handler middleware

## Test Plan

No new tests needed, updating the existing tests gives sufficient coverage.


src/data-access/payIds.ts
```diff
@@ -7,14 +7,55 @@ import { Address, AddressInformation } from '../types/database'
...
+    .select(
+      'address.paymentNetwork',
+      'address.environment',
+      'address.details',
+      'address.identityKeySignature',
+    )
+    .from<Address>('address')
+    .innerJoin('account', 'address.accountId', 'account.id')
+    .where('account.payId', payId)
+    .whereNotNull('address.identityKeySignature')
```
austin-king, line 41: big fan of knex making interfacing with DBs more human readable

src/middlewares/errorHandler.ts
```diff
@@ -40,6 +46,21 @@ export default function errorHandler(
     if (err.kind === ParseErrorType.InvalidMediaType) {
       metrics.recordPayIdLookupBadAcceptHeader()
     }
+
+    // Collect metrics on public API requests to a PayID that does not exist
+    if (
```
dino-rodriguez, line 51: This metrics capturing used to be sitting in the middleware, by tweaking our `LookupError` interface, I was able to put it here alongside the other failure metrics ( e.g. `recordPayIdLookupBadAcceptHeader` ).

This significantly cleaned up the main middleware.

src/middlewares/payIds.ts
```diff
@@ -42,29 +46,36 @@ export default async function getPaymentInfo(
   const parsedAcceptHeaders = parseAcceptHeaders(req.accepts())
 
   // Get all addresses from DB
-  const allAddressInfo = await getAllAddressInfoFromDatabase(payId)
+  const [allAddressInfo] = await Promise.all([
+    getAllAddressInfoFromDatabase(payId),
+    getAllVerifiedAddressInfoFromDatabase(payId),
+    getIdentityKeyFromDatabase(payId).catch((_err) => {
```
dino-rodriguez, line 52: Using `Promise.all` should be significantly more performant, because we run all of these in parallel instead of sequentially with `await`. The catch is that with `Promise.all`, if one promise fails, the whole thing fails, unless you explicitly `catch` the failures.

This actually works really well in this situation, because we can `catch` an identity key failure to determine if a PayID is missing from the database. Previously, this was something we could not do ( we could not distinguish between a missing PayID and missing addresses ). Now we can!

src/middlewares/payIds.ts
```diff
@@ -42,29 +46,36 @@ export default async function getPaymentInfo(
...
+    }),
+  ])
 
   // Content-negotiation to get preferred payment information
   const preferredAddressInfo = getPreferredAddressHeaderPair(
     allAddressInfo,
     parsedAcceptHeaders,
   )
 
-  // TODO:(hbergren) Distinguish between missing PayID in system, and missing address for paymentNetwork/environment.
```
dino-rodriguez, line 53: We were able to handle this above!

src/middlewares/payIds.ts
```diff
@@ -42,29 +46,36 @@ export default async function getPaymentInfo(
...
-      metrics.recordPayIdLookupResult(
-        false,
-        acceptType.paymentNetwork,
-        acceptType.environment,
-      ),
-    )
-
     throw new LookupError(
       `Payment information for ${payId} could not be found.`,
-      LookupErrorType.Unknown,
```
dino-rodriguez, line 67: We were able to move this to the error handler, and also specify a `LookupErrorType` since we disambiguated these two.

src/services/headers.ts
```diff
@@ -8,22 +9,6 @@ const badAcceptHeaderErrorMessage = `Must have an Accept header of the form "app
       - 'Accept: application/payid+json'
       `
 
-/** A parsed HTTP Accept header object. */
```
dino-rodriguez, line 11: Moved this to a dedicated `types` file because having it in here was causing an import cycle.

src/utils/errors/lookupError.ts
```diff
@@ -14,16 +16,23 @@ export enum LookupErrorType {
  */
 export default class LookupError extends PayIDError {
   public readonly kind: LookupErrorType
+  public readonly headers?: readonly ParsedAcceptHeader[]
```
dino-rodriguez, line 19: This new optional parameter allows us to pass in our headers for `LookupErrors`, so that we can then capture them with metrics later in the error middleware.

test/integration/e2e/public-api/basePayId.test.ts
```diff
@@ -151,7 +151,7 @@ describe('E2E - publicAPIRouter - Base PayID', function (): void {
     const expectedErrorResponse = {
       statusCode: 404,
       error: 'Not Found',
-      message: 'Payment information for johndoe$127.0.0.1 could not be found.',
+      message: 'PayID johndoe$127.0.0.1 could not be found.',
```
dino-rodriguez, line 154: Because we can now tell when it's a missing PayID

src/middlewares/payIds.ts
austin-king: me gusta

src/services/headers.ts
austin-king: Cool I'm a fan of that. Makes me realize in my last PR that we define DatabaseAddress and VerifiedDatabaseAddress interfaces in a file containing functions. Would be great if we consolidated all interfaces into our `types` directory so we always know where to go find them.

src/types/headers.ts
```diff
@@ -0,0 +1,15 @@
...
+  /** A raw Accept header media type. */
+  readonly mediaType: string
+
+  /** The payment network requested in the media type. */
+  readonly paymentNetwork: string
+
+  /**
+   * The environment requested in the media type.
+   *
+   * Optional, as some headers (like application/ach+json) don't have an environment.
```
austin-king, line 12: Good comment for context

src/services/headers.ts
dino-rodriguez: Yeah, I agree. Initially I had a discussion with Hans for where types should live, and I wasn't clear on exactly why I thought types should be in a separate place, but had a disposition towards it. I think this is a good case for a concrete reason for why that might be the way to go.

PR #641, opened 2020-08-06
feat: verifiable payid content-negotiation & address filtering

dino-rodriguez:

Updates content-negotiation to support verifiable PayID.

Updates Public API to use content-negotiation for verifiable PayID.

Updates all relevant tests.

### Context of Change

Continuing work on integrating verifiable PayID into the Public API.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Before : Content-negotiation & address filtering of regular addresses
Before : Content-negotiation & address filtering of regular & verified addresses

## Test Plan

Added one new test, updated all old tests. I think we have solid unit test
coverage. After the output payload has been filtered, we will need to update our
existing integration tests, and probably add new ones too.

## Future Tasks

Last part of the public API work is to output verified PayIDs in the payload.


src/middlewares/payIds.ts
```diff
@@ -64,13 +64,14 @@ export default async function getPaymentInfo(
...
-  const preferredAddressInfo = getPreferredAddressHeaderPair(
+  const [preferredHeader, preferredAddresses] = getPreferredAddressHeaderPair(
     allAddressInfo,
+    allVerifiedAddressInfo,
     parsedAcceptHeaders,
   )
 
   // Respond with a 404 if we can't find the requested payment information
-  if (preferredAddressInfo === undefined) {
+  if (!preferredHeader) {
```
dino-rodriguez, line 74: We only want to `404` if the there are no addresses at all ( both verified and unverified ). This header value will only be undefined under that condition.

src/middlewares/payIds.ts
```diff
@@ -64,13 +64,14 @@ export default async function getPaymentInfo(
   ])
 
   // Content-negotiation to get preferred payment information
-  const preferredAddressInfo = getPreferredAddressHeaderPair(
+  const [preferredHeader, preferredAddresses] = getPreferredAddressHeaderPair(
```
dino-rodriguez, line 67: Turning this into a tuple greatly simplified de-structuring verified addresses.

We don't pull out the verified addresses until the next PR because we aren't using them.

src/middlewares/payIds.ts
```diff
@@ -79,11 +80,6 @@ export default async function getPaymentInfo(
     )
   }
 
-  const {
```
dino-rodriguez, line 82: De-structuring this above.

src/services/basePayId.ts
```diff
@@ -44,56 +44,61 @@ export function formatPaymentInfo(
...
+        (address.environment ?? undefined) === acceptHeader.environment,
     )
 
     // Return the address + the media type to respond with
-    if (paymentInformationForAcceptType) {
-      return {
-        preferredParsedAcceptHeader: acceptHeader,
-        preferredAddresses: [paymentInformationForAcceptType],
-      }
+    // If either a unverified or verified address is found, we return
```
dino-rodriguez, line 91: When the first header is matched, we want to return, regardless of whether it yields verified or unverified PayIDs.

This adheres to the content-negotiation preference matching.

PR #648, opened 2020-08-07
test: formatting payment info for v. payid

dino-rodriguez:

Adds test to check that payment info is properly formatted.

### Context of Change

This tests some of the new functionality from the last PR.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Adds a test

## Test Plan

Need to add more integration tests


src/services/basePayId.ts
```diff
@@ -56,7 +56,7 @@ export function formatPaymentInfo(
         },
       }
     }),
-    ...(payId && { payId }),
+    payId,
```
dino-rodriguez, line 59: This is no longer optional because of verifiable PayID.

hbergren: You mean it's no longer an optional parameter, correct?

dino-rodriguez: Yeah that's what I mean, the return value is still optional there - good clarification.

PR #650, opened 2020-08-10
feat: AdminAPI accepts VerifiedAddresses

austin-king:

AdminAPI now accepts a POST containing `VerifiedAddresses`.

### Context of Change

https://app.asana.com/0/1182225689526777/1186929984905885

There is one final PR regarding admin api (PUT functionality) for v. payID. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

AdminAPI now accepts verified addresses when posting a user.

## Test Plan

Previous tests continue to work + I added one confirming it gives us a `201` upon posting a user w/ verified addresses.

## Future Tasks

PUT functionality.
-->

### DIFF
--- a/src/middlewares/users.ts
+++ b/src/middlewares/users.ts
@@ -13,6 +13,7 @@ import {
   replaceUserPayId,
   checkUserExistence,
 } from '../data-access/users'
+import { AddressInformation } from '../types/database'
 import {
   LookupError,
   LookupErrorType,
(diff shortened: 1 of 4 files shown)


src/middlewares/users.ts
```diff
@@ -113,7 +114,16 @@ export async function postUser(
   // TODO:(hbergren) Need to test here and in `putUser()` that `req.body.addresses` is well formed.
   // This includes making sure that everything that is not ACH or ILP is in a CryptoAddressDetails format.
   // And that we `toUpperCase()` paymentNetwork and environment as part of parsing the addresses.
-  await insertUser(payId, req.body.addresses, req.body.identityKey)
+  let allAddresses: AddressInformation[] = []
+
+  if (req.body.addresses !== undefined) {
```
austin-king, line 119: When we put type validation into this middleware these checks can likely be improved

hbergren: What you have is probably fine.

You could also do something like this:
```ts
const allAddresses = [...req.body.addresses || [], ...req.body.verifiedAddresses || []]
```

Typing that out, I think what you have is more readable though haha.

test/integration/e2e/admin-api/postUsers.test.ts
```diff
@@ -48,6 +48,44 @@ describe('E2E - adminApiRouter - POST /users', function (): void {
...
+    // WHEN we make a POST request to /users with that user information
+    request(app.adminApiExpress)
+      .post(`/users`)
+      .set('PayID-API-Version', payIdApiVersion)
+      .send(userInformation)
+      .expect('Content-Type', /text\/plain/u)
+      // THEN we expect the Location header to be set to the path of the created user resource
+      .expect('Location', `/users/${userInformation.payId}`)
+      // THEN we expect back a 201 - CREATED
+      .expect(HttpStatus.Created, done)
```
hbergren, line 86: Is it worth having a subsequent GET request to verify that it the PayID actually saved the `verifiedAddresses`?

src/middlewares/users.ts
austin-king: Ya I like that but I'll just keep it as is

test/integration/e2e/admin-api/postUsers.test.ts
austin-king: Great point, just pushed a commit stringing a GET to confirm afterwards.

PR #651, opened 2020-08-11
test: vpayid protocol integration tests

dino-rodriguez:

Integration tests for the protocol / public API for Verifiable PayID.

### Context of Change

Core integration tests to confirm verifiable PayID protocol works as expected.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Adds tests

## Test Plan

Need to add content-negotiation tests to finish testing, also left some comments
in the PR about adding more integration tests here

### DIFF
--- a/src/db/seed/seeded_values_for_testing.sql
+++ b/src/db/seed/seeded_values_for_testing.sql
@@ -10,6 +10,13 @@ INSERT INTO account(id, pay_id) VALUES
 ('b253bed2-79ce-45d0-bbdd-96867aa85fd5', 'zebra$xpring.money'),
 ('8a75f884-ab16-40c4-a82a-aca454dad6b2', 'empty$xpring.money');
 
+/* Verified accounts*/
+INSERT INTO account(id, pay_id, identity_key) VALUES
+('27944333-faf6-41e8-90c3-1ec9001f0830', 'emptyverified$127.0.0.1', ''),
+('2955cce9-c350-4b60-9726-c415072961ed', 'verified$127.0.0.1', 'bGV0IG1lIHNlZSB0aGVtIGNvcmdpcyBOT1cgb3IgcGF5IHRoZSBwcmljZQ=='),
+('67d9ad5f-5cd8-4a0c-b642-70e63354e647', 'postmalone$127.0.0.1', 'aGkgbXkgbmFtZSBpcyBhdXN0aW4gYW5kIEkgYW0gdGVzdGluZyB0aGluZ3M='),
+('35192b90-9b88-4137-85c9-3d1d3d92cf2c', 'johnwick$127.0.0.1', 'aGV0IG1lIHNlZSB0aGVtIGNvcmdpcyBOT1cgb3IgcGF5IHRoZSBwcmljZQ==');
+
 INSERT INTO address(account_id, payment_network, environment, details) VALUES
 ('232370e9-045e-4269-96ec-5a79091d65ff', 'XRPL', 'MAINNET', '{"address": "rw2ciyaNshpHe7bCHo4bRWq6pqqynnWKQg", "tag": "67298042"}'),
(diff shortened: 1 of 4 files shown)


src/db/seed/seeded_values_for_testing.sql
```diff
@@ -10,6 +10,13 @@ INSERT INTO account(id, pay_id) VALUES
 ('b253bed2-79ce-45d0-bbdd-96867aa85fd5', 'zebra$xpring.money'),
 ('8a75f884-ab16-40c4-a82a-aca454dad6b2', 'empty$xpring.money');
 
+/* Verified accounts*/
```
dino-rodriguez, line 13: I re-ordered these seeds because I want to test adding both verified & unverified addresses to an account, and before since the unverified addresses came before the verified accounts, I didn't know if it would work.

src/db/seed/seeded_values_for_testing.sql
```diff
@@ -18,15 +25,15 @@ INSERT INTO address(account_id, payment_network, environment, details) VALUES
 ('223ece9c-2a15-48e1-9df6-d9ac77c5db90', 'XRPL', 'TESTNET', '{"address": "rDk7FQvkQxQQNGTtfM2Fr66s7Nm3k87vdS"}'),
 ('ec06236e-d134-4a7b-b69e-0606fb54b67b', 'XRPL', 'TESTNET', '{"address": "rDk7FQvkQxQQNGTtfM2Fr66s7Nm3k87vdS"}'),
 ('69b0d20a-cdef-4bb9-adf9-2109979a12af', 'XRPL', 'TESTNET', '{"address": "rDk7FQvkQxQQNGTtfM2Fr66s7Nm3k87vdS"}'),
+/*('35192b90-9b88-4137-85c9-3d1d3d92cf2c', 'BTC',  'TESTNET', '{"address": "2NGZrVvZG92qGYqzTLjCAewvPZ7JE8S8VxE"}'),*/
```
dino-rodriguez, line 28: This is in reference to the above comment -- I want to test retrieving from an account with both unverified and verified addresses. However after  building / running the DB migration with this, and running tests, the tests fail with no error message ( like they do sometimes when there is something wrong with the DB ).

@hbergren @austin-king Is this because I am trying to add 2 of the same type of account ( BTC TESTNET ) into the addresses table?

hbergren: Yeah, there is a unique constraint on (paymentNetwork, environment, accountId), so any given account can only have one BTC TESTNET address. There is another BTC TESTNET address for this account added below.

test/integration/e2e/admin-api/metrics.test.ts
```diff
@@ -100,7 +100,7 @@ describe('E2E - adminApiRouter - GET /metrics', function (): void {
 
     // We create 8 PayIDs in the tests before this one,
     // and start with 5 seeded PayIDs, for a total of 13.
-    await assertMetrics(/actual_payid_count\{org="127.0.0.1"\} 17/u)
+    await assertMetrics(/actual_payid_count\{org="127.0.0.1"\} 18/u)
```
hbergren, line 103: Probably update the comment above this as well.

test/integration/e2e/public-api/verifiablePayId.test.ts
```diff
@@ -0,0 +1,230 @@
...
+            },
+          ],
+        },
+      ],
+    }
+
+    // WHEN we make a GET request to the public endpoint to retrieve payment info with an Accept header specifying xrpl-mainnet
+    request(app.publicApiExpress)
+      .get(payId)
+      .set('PayID-Version', '1.0')
```
hbergren, line 58: I think this is probably fine, but are we incrementing the version of the PayID Protocol with V. PayID?

dino-rodriguez: Yeah. Even though it doesn't affect those tests, probably makes sense to use the latest version for all of them.

PR #652, opened 2020-08-11
test: vpayid content-negotiation tests

dino-rodriguez:

Adds integration tests for content-negotiation functionality of verifiable
PayID.

### Context of Change

Final set of integration tests for initial v. payid integration in the public
api.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Adds new tests


test/integration/e2e/public-api/basePayIdContentNegotiation.test.ts
```diff
@@ -0,0 +1,172 @@
...
+const XRPL_TESTNET_ACCEPT_HEADER = 'application/xrpl-testnet+json'
+const XRPL_MAINNET_ACCEPT_HEADER = 'application/xrpl-mainnet+json'
+
+describe('E2E - publicAPIRouter - Content Negotiation', function (): void {
+  // Boot up Express application and initialize DB connection pool
+  before(async function () {
+    app = await appSetup()
+  })
+
+  it('Returns the first address for multiple types with no q', function (done): void {
```
florent-uzio, line 21: @dino-rodriguez , does `q` mean `queue` here (and below)?

test/integration/e2e/public-api/basePayIdContentNegotiation.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      ],
+      verifiedAddresses: [],
+      payId: 'alice$127.0.0.1',
+    }
+
+    // WHEN we make a GET request to the public endpoint to retrieve payment info with an Accept header specifying
+    // both testnet and mainnet, with no q for either
+    request(app.publicApiExpress)
+      .get(payId)
+      .set('PayID-Version', '1.0')
```
florent-uzio, line 44: @dino-rodriguez , why do we have `1.0` in this file whereas the version is `1.1` in `test/integration/e2e/public-api/verifiablePayIdContentNegotiation.test.ts` ?

florent-uzio:
Nice job @dino-rodriguez !
What should happen if `q` is the same for both XRPL mainnet and testnet for example?

I did a test and only the first `environment` in the `Accept` header is returned. Is it what's expected or shall we return both `mainnet` and `testnet` in that case?

image

test/integration/e2e/public-api/basePayIdContentNegotiation.test.ts
dino-rodriguez: Hey Florent, good question! `q` in this case stands for `q-factor-weighting`, which is used to explicitly specify the order of preference for `Accept` headers when using content-negotiation.

For more information, I would suggest taking a look here : https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

test/integration/e2e/public-api/basePayIdContentNegotiation.test.ts
dino-rodriguez: Good catch, I will update that.

dino-rodriguez:
Thanks @florent-uzio !

In the case of the same `q` weight, I believe the specificity of the header & the order take precedence. And yeah, the expected behavior is that one should be returned.

test/integration/e2e/public-api/verifiablePayIdContentNegotiation.test.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  it('Returns the preferred available address where the higher q is at the beginning', function (done): void {
+    // GIVEN a payment pointer known to have an associated xrpl-testnet address and xrpl-mainnet address
+    const payId = '/johnwick'
+    const acceptHeader = `${XRPL_TESTNET_ACCEPT_HEADER}; q=1.1, ${XRPL_MAINNET_ACCEPT_HEADER}; q=0.5`
+    const expectedResponse = {
+      addresses: [],
+      verifiedAddresses: [
+        {
+          signatures: [
```
theotherian, line 77: Same question.

test/integration/e2e/public-api/verifiablePayIdContentNegotiation.test.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  it('Returns the preferred available address where the higher q is at the end', function (done): void {
+    // GIVEN a payment pointer known to have an associated xrpl-testnet address and an xrpl-mainnet address
+    const payId = '/johnwick'
+    const acceptHeader = `${XRPL_TESTNET_ACCEPT_HEADER}; q=0.5, ${XRPL_MAINNET_ACCEPT_HEADER}; q=1.1`
+    const expectedResponse = {
+      addresses: [],
+      verifiedAddresses: [
+        {
+          signatures: [
```
theotherian, line 125: Same question.

test/integration/e2e/public-api/verifiablePayIdContentNegotiation.test.ts
```diff
@@ -0,0 +1,218 @@
...
+  it('Returns a valid address when the most preferred type does not exist', function (done): void {
+    // GIVEN a payment pointer known to have an associated xrpl-testnet address and mainnet address
+    const payId = '/johnwick'
+    const nonExistentAcceptType = 'application/fakenetwork-fakenet+json'
+    const acceptHeader = `${nonExistentAcceptType}; q=1.1, ${XRPL_TESTNET_ACCEPT_HEADER}; q=0.5, ${XRPL_MAINNET_ACCEPT_HEADER}; q=0.9`
+    const expectedResponse = {
+      addresses: [],
+      verifiedAddresses: [
+        {
+          signatures: [
```
theotherian, line 174: Same question.

test/integration/e2e/public-api/basePayIdContentNegotiation.test.ts
```diff
@@ -0,0 +1,172 @@
...
+  // Boot up Express application and initialize DB connection pool
+  before(async function () {
+    app = await appSetup()
+  })
+
+  it('Returns the first address for multiple types with no q', function (done): void {
+    // GIVEN a payment pointer known to have an associated xrpl-testnet and xrpl-mainnet address
+    const payId = '/alice'
+    const acceptHeader = `${XRPL_TESTNET_ACCEPT_HEADER}, ${XRPL_MAINNET_ACCEPT_HEADER}`
+    const expectedResponse = {
```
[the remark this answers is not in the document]
dino-rodriguez: This technically wasn't really part of this PR ( I renamed this file just so that I could make the relevant one for verifiable PayID ), but I think it's a good refactor since I'm already touching these files so I went ahead and did it.

test/integration/e2e/public-api/basePayIdContentNegotiation.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      })
+      // AND we get back the xrpl testnet account information associated with that payment pointer for xrpl testnet
+      .expect(HttpStatus.OK, expectedResponse, done)
+  })
+
+  it('Returns the preferred available address where the higher q is at the end', function (done): void {
+    // GIVEN a payment pointer known to have an associated xrpl-testnet address and an xrpl-mainnet address
+    const payId = '/alice'
+    const acceptHeader = `${XRPL_TESTNET_ACCEPT_HEADER}; q=0.5, ${XRPL_MAINNET_ACCEPT_HEADER}; q=1.1`
+    const expectedResponse = {
```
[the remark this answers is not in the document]
dino-rodriguez: Extracting this as well. For now I'm going to leave it in this file, as it is only used here and not in the verifiable PayID content-negotiation tests.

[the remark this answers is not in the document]
dino-rodriguez: Yeah, I think extracting to an external file to share amongst these is a good idea.

Hopefully down the line we can cut out most of the redundant json payloads in tests.

test/integration/e2e/public-api/basePayIdContentNegotiation.test.ts
```diff
@@ -7,10 +7,27 @@ import * as request from 'supertest'
...
+  XRPL_MAINNET_ADDRESS,
+  XRPL_TESTNET_ACCEPT_HEADER,
+  XRPL_MAINNET_ACCEPT_HEADER,
+} from './payloads'
+
 let app: App
 
-const XRPL_TESTNET_ACCEPT_HEADER = 'application/xrpl-testnet+json'
-const XRPL_MAINNET_ACCEPT_HEADER = 'application/xrpl-mainnet+json'
+const USER = '/alice'
```
dino-rodriguez, line 19: Refactoring this to constants to reduce repeated code.

test/integration/e2e/public-api/payloads.ts
```diff
@@ -0,0 +1,30 @@
+import { Address, AddressDetailsType } from '../../../../src/types/protocol'
```
dino-rodriguez, line 1: Defining some constants here that get re-used in the conneg tests.

They may also be re-used elsewhere, so maybe we can continue to expand on this down the line.

test/integration/e2e/public-api/basePayIdContentNegotiation.test.ts
florent-uzio: Hey Dino, thank you. Do you think it would make sense to include that link in the code for people who are not familiar with it (like me)?

PR #653, opened 2020-08-12
Set payid metrics protocol version and server agent

nhartner:

Updates the metrics config to set the PayID protocol version and the server agent

### Context of Change

This will allow the server and protocol versions for people running the reference implementation to be tracked within the metrics platform

### Type of Change

- [ X] New feature (non-breaking change which adds functionality)
- [ ] Documentation Updates
- [ ] Release

## Before / After

Just updated the config settings

## Test Plan

Run PayID server and verify that the new metrics are set and published to the Xpring metrics infra.


package.json
```diff
@@ -40,7 +40,7 @@
   },
   "dependencies": {
     "@hapi/boom": "^9.1.0",
-    "@payid-org/server-metrics": "^1.0.0",
+    "@payid-org/server-metrics": "github:payid-org/server-metrics#nh/server-and-protocol-version",
```
nhartner, line 43: FIXME: I need to cut a new release of server-metrics and fix this dependency before merging.

src/config.ts
```diff
@@ -59,6 +59,12 @@ const config = {
     /** How frequently (in seconds) to refresh the PayID Count report data from the database. */
     payIdCountRefreshIntervalInSeconds:
       Number(process.env.PAYID_COUNT_REFRESH_INTERVAL) || 60,
+
+    payIdProtocolVersion: payIdServerVersions[payIdServerVersions.length - 1],
+
+    serverAgent: `@payid-org/payid:${
+      process.env.npm_package_version ?? 'unknown.version'
```
dino-rodriguez, line 66: Is this variable always in the environment when a node app is started?

hbergren: Echoing Dino, I assume you've tested this, right?

dino-rodriguez:
Whoops, I just accidentally re-requested @hbergren for a review on this PR, meant to do it on the original metrics PR.

src/config.ts
nhartner: Yeah, good question. I only tested it using `npm start` which works, but it doesn't work inside the docker container when started using `node ...`.

nhartner: Pushed a change to read this from package.json instead which I verified works using both npm and node to run the server.

package.json
```diff
@@ -16,12 +16,13 @@
...
     "build": "run-script-os",
-    "build:default": "rm -rf build && tsc --project . && npm run copySQL && npm run copyHTML && npm run copyJSON",
-    "build:win32": "del /q /f build && tsc --project . && npm run copySQL && npm run copyHTML && npm run copyJSON",
+    "build:default": "rm -rf build && tsc --project . && npm run copySQL && npm run copyHTML && npm run copyJSON && npm run copyPackageJSON",
+    "build:win32": "del /q /f build && tsc --project . && npm run copySQL && npm run copyHTML && npm run copyJSON && npm run copyPackageJSON",
     "buildWatch": "tsc --watch --project .",
     "copyHTML": "copyfiles -u 1 src/html/* build/src",
     "copySQL": "copyfiles -u 1 **/*.sql build/src",
     "copyJSON": "copyfiles -u 1 src/**/*.json build/src",
+    "copyPackageJSON": "copyfiles package.json build",
```
hbergren, line 25: Really clever solution.

PR #661, opened 2020-08-17
fix: Verifiable PayID JWS Consumable

dino-rodriguez:

Outputs verifiable PayID payloads as stringified JSON, instead of JSON.

### Context of Change

After discussing with @nhartner, I learned that the `payload` needs to be
stringified JSON, so that it can be verified as a JWS ( as per RFC7797 ).

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Updates protocol output, and relevant tests

## Test Plan

No new tests needed


src/types/protocol.ts
```diff
@@ -49,7 +49,7 @@ export interface Address {
  * Object containing address information alongside signatures.
  */
 interface VerifiedAddress {
-  readonly payload: VerifiedAddressPayload
+  readonly payload: VerifiedAddressPayload | string
```
nhartner, line 52: Do we still need to keep this as `VerifiedAddressPayLoad | string` vs just `string`?

austin-king: think this might have been a leftover from the last minute changes to make the demo work

PR #662, opened 2020-08-17
Add version to payment information

hbergren:

Adds a `version` key to the PaymentInformation response payload.

### Context of Change

This is needed to cut the `PayID-Version` and `PayID-Server-Version` headers from the PayID Protocol when we hit 2.0. I spoke with Aanchal about versioning, and she and I came to the conclusion that (for the moment) we don't even need version negotiation, so we don't need to build out versioning in the PayID Discovery WebFinger JRD like I suggested. You can get all the information you need to know from the `version` key in the payload itself, and (hopefully) there shouldn't be many major version changes for the foreseeable future after we cut 2.0.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

The PaymentInformation response payload now includes a `version` key of the form `version: MAJOR.MINOR`.

## Test Plan

Tests have been updated and all tests pass.


test/integration/e2e/public-api/basePayIdContentNegotiation.test.ts
```diff
@@ -18,15 +18,19 @@ let app: App
 
 const USER = '/alice'
 const PAYID = `${USER.slice(1)}$127.0.0.1`
+const VERSION = '1.1'
```
austin-king, line 21: Can we establish a global const like this one in the rest of the test files in this PR so that we are always using `VERSION` instead of typing out `'1.1'` for each?

dino-rodriguez:
Actually one comment - I noticed that you mentioned we don't need version negotiation. That means that when we go from 1.x to 2.0, we are going to break the network, right?

test/integration/e2e/public-api/basePayIdContentNegotiation.test.ts
dino-rodriguez: Any chance we can do this refactor in a subsequent PR so we can get this release cut?

PR #672, opened 2020-08-24
Patch API update / fix for V. PayID

hbergren:

Fix the PATCH API now that V. PayID exists.

### Context of Change

If you update the PayID of an account that has associated signed addresses, you would break the signatures of those addresses.

So, instead of letting you PATCH the PayID of an account, we now throw an error if any addresses on that account have an associated identity signature.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added a test for this case, and all tests pass.

### DIFF
--- a/src/routes/publicApiRouter.ts
+++ b/src/routes/publicApiRouter.ts
@@ -18,7 +18,7 @@ publicApiRouter
   // Allow the PayID Protocol to basically ignore CORS
   .use((_req, res, next) => {
     res.header('Access-Control-Allow-Origin', '*')
-    res.header('Access-Control-Allow-Methods', 'POST, GET, OPTIONS')
+    res.header('Access-Control-Allow-Methods', 'GET, OPTIONS')
     res.header('Access-Control-Allow-Headers', 'PayID-Version')
     res.header(
       'Access-Control-Expose-Headers',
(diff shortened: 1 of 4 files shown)


src/routes/publicApiRouter.ts
```diff
@@ -18,7 +18,7 @@ publicApiRouter
   // Allow the PayID Protocol to basically ignore CORS
   .use((_req, res, next) => {
     res.header('Access-Control-Allow-Origin', '*')
-    res.header('Access-Control-Allow-Methods', 'POST, GET, OPTIONS')
+    res.header('Access-Control-Allow-Methods', 'GET, OPTIONS')
```
hbergren, line 21: Unrelated, but `POST` should not be in this header, and the PayID Validator now has an assertion for this.

src/utils/errors/parseError.ts
```diff
@@ -21,6 +21,9 @@ export enum ParseErrorType {
   MissingPayIdApiVersionHeader = 'MissingPayIdApiVersionHeader',
   InvalidPayIdApiVersionHeader = 'InvalidPayIdApiVersionHeader',
   UnsupportedPayIdApiVersionHeader = 'UnsupportedPayIdApiVersionHeader',
+
+  // Verifiable PayID Stuff
+  IncompatibleRequestMethod = 'IncompatibleRequestMethod',
```
hbergren, line 26: Not sure what to call this error type, but "Using a PATCH when you should only use a PUT" seemed to be an "IncompatibleRequestMethod".

PR #674, opened 2020-08-25
feat: Admin API Matches Public API

dino-rodriguez:

Adds a new Admin API version.

If using the new version, the Admin API POST accepts a format equivalent to the
Public API.

It checks to make sure that only one identity key is used amongst all the
signatures.

### Context of Change

The first of a sequence of changes to conform the Admin API format to the public
API.

POST validates that only a single identity key is used, but more work needs to
be done for POST to actually work on the new version. This is the first PR.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

After : Admin API POST accepts Public API format on new version, but currently
does not actually post ( only validates identity keys )

### DIFF
--- a/src/middlewares/users.ts
+++ b/src/middlewares/users.ts
@@ -140,8 +140,9 @@ export async function postUser(
   // If using "new" ( same as Public API ) API format
   else if (requestVersion >= adminApiVersions[1]) {
     req.body.verifiedAddresses.forEach((address: VerifiedAddress) => {
+      let identityKeyCount = 0
+
       address.signatures.forEach((signature: VerifiedAddressSignature) => {
-        let identityKeyCount = 0
         const decodedKey = JSON.parse(
           Buffer.from(signature.protected, 'base64').toString(),
         )
--- a/src/middlewares/users.ts
+++ b/src/middlewares/users.ts
@@ -120,19 +120,20 @@ export async function postUser(
   // And that we `toUpperCase()` paymentNetwork and environment as part of parsing the addresses.
   let allAddresses: AddressInformation[] = []
   let identityKey: string | undefined
+  // This is defined because we checked it in checkRequestAdminApiVersionHeaders middleware
+  const requestVersion = String(req.get('PayID-API-Version'))
 
   if (req.body.addresses !== undefined) {
     allAddresses = allAddresses.concat(req.body.addresses)
   }
   // Checking for existence of identity key to see if we are using old Admin API format
-  // TODO(dino): Use version instead of object inspection
   if (
     req.body.verifiedAddresses !== undefined &&
-    res.get('PayID-API-Version') < adminApiVersions[1]
+    requestVersion < adminApiVersions[1]
   ) {
     allAddresses = allAddresses.concat(req.body.verifiedAddresses)
     identityKey = req.body.identityKey
-  } else if (res.get('PayID-API-Version') >= adminApiVersions[1]) {
+  } else if (requestVersion >= adminApiVersions[1]) {
     req.body.verifiedAddresses.forEach((address: VerifiedAddress) => {
       address.signatures.forEach((signature: VerifiedAddressSignature) => {
         const decodedKey = JSON.parse(
--- a/src/middlewares/users.ts
+++ b/src/middlewares/users.ts
@@ -97,6 +97,7 @@ export async function getUser(
  *
  * @throws ParseError if the PayID is missing from the request body.
  */
+// eslint-disable-next-line max-lines-per-function -- Disabling until I finish building the functionality here.
 export async function postUser(
   req: Request,
   res: Response,
@@ -117,6 +118,8 @@ export async function postUser(
 
   // We can be sure the version is defined because we verified it in checkRequestAdminApiVersionHeaders middleware
   const requestVersion = String(req.get('PayID-API-Version'))
+  const identityKeyLabel = 'identityKey'
+
   let allAddresses: AddressInformation[] = []
   let identityKey: string | undefined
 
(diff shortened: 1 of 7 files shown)


src/types/protocol.ts
```diff
@@ -48,16 +48,16 @@ export interface Address {
...
   readonly signatures: readonly VerifiedAddressSignature[]
 }
 
 /**
  * JWS object for verification.
  */
-interface VerifiedAddressSignature {
-  name: string
+export interface VerifiedAddressSignature {
+  name?: string
```
dino-rodriguez, line 60: Separate fix, but this should be an optional field since it is not defined in the spec.

src/middlewares/users.ts
```diff
@@ -113,19 +115,51 @@ export async function postUser(
...
+  // Check version header to determine which Admin API format we are using
+  if (
+    req.body.verifiedAddresses !== undefined &&
+    requestVersion < adminApiVersions[1]
+  ) {
     allAddresses = allAddresses.concat(req.body.verifiedAddresses)
+    identityKey = req.body.identityKey
+  } else if (requestVersion >= adminApiVersions[1]) {
+    req.body.verifiedAddresses.forEach((address: VerifiedAddress) => {
+      address.signatures.forEach((signature: VerifiedAddressSignature) => {
```
dino-rodriguez, line 138: Takes the first identity key found, and goes through all signature arrays to verify that the identity keys are all the same.

src/middlewares/users.ts
```diff
@@ -113,19 +115,52 @@ export async function postUser(
...
+          )
+        }
+        // Transform to format consumable by insert user
+        // TODO(dino): Implement format translation
+      })
+    })
   }
 
-  await insertUser(payId, allAddresses, req.body.identityKey)
+  // TODO(dino): Verify that the identity key matches the one in the database. Is there a knex way to do this?
```
dino-rodriguez, line 162: @hbergren Do you know if there is a good way to tackle this without having to make a DB call?

src/utils/errors/parseError.ts
```diff
@@ -12,6 +12,9 @@ export enum ParseErrorType {
   MissingPayId = 'MissingPayId',
   InvalidPayId = 'InvalidPayId',
 
+  // Verifiable PayID stuff
+  MultipleIdentityKeys = 'MultipleIdentityKeys',
```
dino-rodriguez, line 16: New error.

src/middlewares/users.ts
```diff
@@ -113,19 +115,52 @@ export async function postUser(
...
   }
-  if (req.body.verifiedAddresses !== undefined) {
+  // Check version header to determine which Admin API format we are using
+  if (
+    req.body.verifiedAddresses !== undefined &&
+    requestVersion < adminApiVersions[1]
+  ) {
     allAddresses = allAddresses.concat(req.body.verifiedAddresses)
+    identityKey = req.body.identityKey
+  } else if (requestVersion >= adminApiVersions[1]) {
```
dino-rodriguez, line 136: Should we check this with the version header, or the one in the body of the payload ( or potentially and OR of both ) ?

src/middlewares/users.ts
```diff
@@ -113,19 +116,73 @@ export async function postUser(
...
+          identityKey = signature.protected
+          identityKeyCount += 1
+        } else {
+          // Increment the count of identity keys per address
+          if (decodedKey.name === identityKeyLabel) {
+            identityKeyCount += 1
+          }
+
+          // Identity key must match across all addresses
+          if (
```
dino-rodriguez, line 160: There are 2 different checks :

- The identity should be the same across all addresses
- There should only be one identity key per address

src/types/protocol.ts
hbergren: I... think I want to make this required?

We could also update the spec, right?

src/middlewares/users.ts
```diff
@@ -95,6 +97,7 @@ export async function getUser(
  *
  * @throws ParseError if the PayID is missing from the request body.
  */
+// eslint-disable-next-line max-lines-per-function -- Disabling until I finish building the functionality here.
```
austin-king, line 100: Did you leave this so it didn't give you an error while building and meant to remove or is there a plan to remove this later?

src/middlewares/users.ts
dino-rodriguez: Ok, that makes sense

src/middlewares/users.ts
dino-rodriguez: Leaving this in until I finish the whole POST function, and then I intend to refactor into a helper when I write PUT.

src/middlewares/users.ts
```diff
@@ -113,19 +116,73 @@ export async function postUser(
...
+  else if (requestVersion >= adminApiVersions[1]) {
+    req.body.verifiedAddresses.forEach((address: VerifiedAddress) => {
+      address.signatures.forEach((signature: VerifiedAddressSignature) => {
+        let identityKeyCount = 0
+        const decodedKey = JSON.parse(
+          Buffer.from(signature.protected, 'base64').toString(),
+        )
+
+        // Get the first identity key
+        if (!identityKey && decodedKey.name === identityKeyLabel) {
```
hbergren, line 150: Just to double-check, `decodedKey.name` is the `name` property on the _inside_ of the base64 encoded `protected` header, correct?

dino-rodriguez: Correct.

src/types/protocol.ts
dino-rodriguez: I'm down for it to be required if we want that. Can leave it as such here in that case.

src/middlewares/users.ts
dino-rodriguez: Removed here : https://github.com/payid-org/payid/pull/677

src/middlewares/users.ts
```diff
@@ -113,19 +116,73 @@ export async function postUser(
...
+            decodedKey.name === identityKeyLabel
+          ) {
+            throw new ParseError(
+              'More than one identity key detected. Only one identity key per PayID can be used.',
+              ParseErrorType.MultipleIdentityKeys,
+            )
+          }
+
+          // Each address must have only one identity key / signature pair
+          if (identityKeyCount > 1) {
```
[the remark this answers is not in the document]
dino-rodriguez: Ok, yeah - now that I've got something working with this PR & https://github.com/payid-org/payid/pull/677 , I'm going to go ahead and try to tackle a refactor in a 3rd PR on this stack so I have everything to work with for the refactor.

Going to see if I can do this in a more functional way instead of needing to have stateful for loops.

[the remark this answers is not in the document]
dino-rodriguez: I'm not planning on merging this PR though, even after it gets approved because I only want to merge when I am sure Austin's tests are passing on it.

src/types/protocol.ts
hbergren: Yeah, I guess I'm ok with leaving it optional.

src/middlewares/users.ts
austin-king: got it, didn't notice this was a multi part PR

src/middlewares/users.ts
```diff
@@ -113,19 +116,74 @@ export async function postUser(
...
+    req.body.verifiedAddresses !== undefined &&
+    requestVersion < adminApiVersions[1]
+  ) {
     allAddresses = allAddresses.concat(req.body.verifiedAddresses)
+    identityKey = req.body.identityKey
+  }
+  // If using "new" ( same as Public API ) API format
+  else if (requestVersion >= adminApiVersions[1]) {
+    req.body.verifiedAddresses.forEach((address: VerifiedAddress) => {
+      let identityKeyCount = 0
```
dino-rodriguez, line 143: I had a bug where this was inside the `signatures` for loop instead of `addresses` - with this fixed, is there still a way I can drop count from here that I'm not seeing @hbergren ?

PR #677, opened 2020-08-26
[2/x] Admin API Matches Public API: POST - Insert Verified User

dino-rodriguez:

POST translates from the new format to one that is acceptable by insert user.

### Context of Change

This is necessary so that POST works with our existing DB queries.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Befor : POST with the new format only checked identity keys
After : POST with the new format now inserts

## Test Plan

Austin is tackling tests in other PRs.

### DIFF
--- a/src/middlewares/users.ts
+++ b/src/middlewares/users.ts
@@ -140,20 +140,25 @@ export async function postUser(
   // If using "new" ( same as Public API ) API format
   else if (requestVersion >= adminApiVersions[1]) {
     req.body.verifiedAddresses.forEach((address: VerifiedAddress) => {
+      let identityKeySignature: string | undefined
+
       address.signatures.forEach((signature: VerifiedAddressSignature) => {
         let identityKeyCount = 0
         const decodedKey = JSON.parse(
           Buffer.from(signature.protected, 'base64').toString(),
         )
 
-        // Get the first identity key
+        // Get the first identity key & signature
         if (!identityKey && decodedKey.name === identityKeyLabel) {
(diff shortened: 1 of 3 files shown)


src/middlewares/users.ts
```diff
@@ -175,13 +180,22 @@ export async function postUser(
...
+        details: {
+          address: jwsPayload.payIdAddress.addressDetails.address,
+        },
+        identityKeySignature,
+      }
+      allAddresses.push(addressPayload)
     })
   }
 
-  // TODO(dino): Verify that the identity key matches the one in the database. Is there a knex way to do this?
```
dino-rodriguez, line 184: We actually didn't need to do this, because POST is only for creating a new PayID ( so by definition there is no identity key ).

src/middlewares/users.ts
```diff
@@ -140,20 +140,25 @@ export async function postUser(
...
         const decodedKey = JSON.parse(
           Buffer.from(signature.protected, 'base64').toString(),
         )
 
-        // Get the first identity key
+        // Get the first identity key & signature
         if (!identityKey && decodedKey.name === identityKeyLabel) {
           identityKey = signature.protected
           identityKeyCount += 1
+          identityKeySignature = signature.signature
```
austin-king, line 155: signature -> verifiedPayload or something so that we can have verifiedPayload.signature instead of signature.signature?

dino-rodriguez: Sure, good call

PR #678, opened 2020-08-27
[3/x] Admin API Matches Public API: POST - Verify Identity Key + Insert Verified User ( Refactor )

dino-rodriguez:

Pulls out the work in the Admin API POST from the previous 2 PRs in the stack
into a helper : `parseVerifiedAddresses`.

### Context of Change

This will allow this code to be used for `PUT`, and it can also be tested.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Pulls out code into a helper

## Test Plan

Need to write integration tests, and maybe also unit tests. If we can get full
test coverage using integration tests, I don't think we really need unit tests
that duplicate the same logic.

### DIFF
--- a/src/middlewares/users.ts
+++ b/src/middlewares/users.ts
@@ -16,8 +16,7 @@ import {
   replaceUserPayId,
   checkUserExistence,
 } from '../data-access/users'
-import parseVerifiedAddresses from '../services/users'
-import { AddressInformation } from '../types/database'
+import parseAllAddresses from '../services/users'
 import {
   LookupError,
   LookupErrorType,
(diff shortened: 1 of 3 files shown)


src/middlewares/users.ts
```diff
@@ -139,63 +137,9 @@ export async function postUser(
...
-        details: {
-          address: jwsPayload.payIdAddress.addressDetails.address,
-        },
-        identityKeySignature,
-      }
-      allAddresses.push(databaseAddressPayload)
-    })
+    const addressesAndKey = parseVerifiedAddresses(req.body.verifiedAddresses)
+    identityKey = addressesAndKey[1]
+    allAddresses = allAddresses.concat(addressesAndKey[0])
```
hbergren, line 142:
```suggestion
    const [verifiedAddresses, identityKey] = parseVerifiedAddresses(req.body.verifiedAddresses)
    allAddresses = allAddresses.concat(verifiedAddresses)
```

???

dino-rodriguez: We can't do this because `identityKey` needs to be a `let` since it might be assigned if someone is using the old API version : https://github.com/payid-org/payid/pull/678/files#diff-1b08a9ec204937c71764f75cd585f789R136

austin-king: Can we just have both bet `let` then? Think `let [verifiedAddresses, identityKey]` is cleaner / more readable than the alternative

dino-rodriguez: That unfortunately won't work either because `let` is block scoped, so it wouldn't apply outside of this if statement.

PR #679, opened 2020-08-27
[4/x] Admin API Matches Public API : PUT - Verify Identity Key + Insert Verified User

dino-rodriguez:

Parses & PUTS verified addresses using the canonical format.

### Context of Change

Part of all the format changes.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

After : Now puts verified addresses.

## Test Plan

Needs tests

### DIFF
--- a/src/middlewares/users.ts
+++ b/src/middlewares/users.ts
@@ -211,14 +211,18 @@ export async function putUser(
   const requestVersion = String(req.get('PayID-API-Version'))
   const payId = rawPayId.toLowerCase()
   const newPayId = rawNewPayId.toLowerCase()
+
   let identityKey: string | undefined
+  let allAddresses: AddressInformation[] = []
 
   // TODO:(dino) validate body params before this
-  let allAddresses: AddressInformation[] = []
   if (req.body.addresses !== undefined) {
     allAddresses = allAddresses.concat(req.body.addresses)
   }
-  if (req.body.verifiedAddresses !== undefined) {
+  if (
+    req.body.verifiedAddresses !== undefined &&
+    requestVersion < adminApiVersions[1]
+  ) {
     allAddresses = allAddresses.concat(req.body.verifiedAddresses)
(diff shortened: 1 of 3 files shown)


src/middlewares/users.ts
```diff
@@ -113,10 +113,9 @@ export async function postUser(
     )
   }
 
-  const payId = rawPayId.toLowerCase()
-
   // We can be sure the version is defined because we verified it in checkRequestAdminApiVersionHeaders middleware
   const requestVersion = String(req.get('PayID-API-Version'))
+  const payId = rawPayId.toLowerCase()
```
dino-rodriguez, line 118: Just managing whitespace here.

src/middlewares/users.ts
```diff
@@ -171,7 +170,6 @@ export async function putUser(
   // TODO(hbergren): pull this PayID / HttpError out into middleware?
   const rawPayId = req.params.payId
   const rawNewPayId = req.body?.payId
-  const identityKey = req.body?.identityKey
```
dino-rodriguez, line 174: This shouldn't be a const because it can be re-assigned.

src/middlewares/users.ts
```diff
@@ -247,12 +262,15 @@ export async function putUser(
     Boolean(address.identityKeySignature),
   )
 
-  res.locals.status = statusCode
-  res.locals.response = {
-    payId: newPayId,
-    addresses,
-    verifiedAddresses,
+  // Only show created output on the old version
+  if (requestVersion < adminApiVersions[1]) {
```
dino-rodriguez, line 266: Since PUT just deletes all existing addresses, the concept of returning the “updated addresses” really just always returns the same payload the user sent ( since “updated addresses” = “all addresses” ). Therefore in this new version, we are just cutting out the logic of returning the payload instead of complicating this.

PR #683, opened 2020-08-28
[6/x] Admin API Matches Public API: PUT/POST - parseAllAddresses (Refactor + Feature)

dino-rodriguez:

Parses un-verified addresses that use the Public API format for both PUT & POST.

Refactors all of the parsing logic into a separate helper so that it can be
re-used in PUT and POST.

Some minor changes to simplify some of the logic at the end of PUT.

### Context of Change

This is the last change to update PUT & POST to use the Public API format.

### Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Updates
- [ ] Release

## Before / After

After : PUT & POST fully on the new format if using the new version header

## Test Plan

Austin already wrote tests for this

### DIFF
--- a/src/middlewares/users.ts
+++ b/src/middlewares/users.ts
@@ -16,9 +16,7 @@ import {
   replaceUserPayId,
   checkUserExistence,
 } from '../data-access/users'
-import parseVerifiedAddresses from '../services/users'
-import { AddressInformation } from '../types/database'
-import { Address } from '../types/protocol'
+import parseAllAddresses from '../services/users'
 import {
   LookupError,
   LookupErrorType,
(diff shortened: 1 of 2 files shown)


src/middlewares/users.ts
```diff
@@ -222,69 +193,51 @@ export async function putUser(
...
     identityKey,
   )
+  // If user does not exist, create
   if (updatedAddresses === null) {
     updatedAddresses = await insertUser(newPayId, allAddresses, identityKey)
-    statusCode = HttpStatus.Created
-  }
-
-  // If the status code is 201 - Created, we need to set a Location header later with the PayID
-  if (statusCode === HttpStatus.Created) {
```
dino-rodriguez, line 259: We don't need to check this explicitly - if `insertUser` doesn't throw, that means we created a user.

src/middlewares/users.ts
```diff
@@ -222,69 +193,51 @@ export async function putUser(
...
+    const verifiedAddresses = updatedAddresses.filter((address) =>
+      Boolean(address.identityKeySignature),
+    )
     res.locals.response = {
       payId: newPayId,
       addresses,
       verifiedAddresses,
     }
   }
-  res.locals.status = statusCode
```
dino-rodriguez, line 283: Moved up above.

src/middlewares/users.ts
```diff
@@ -222,69 +193,51 @@ export async function putUser(
...
-      details: address.details,
-    }))
-
-  const verifiedAddresses = updatedAddresses.filter((address) =>
-    Boolean(address.identityKeySignature),
-  )
-
-  // Only show created output on the old version
+  // Only show created output on the "old" API
   if (requestVersion < adminApiVersions[1]) {
```
dino-rodriguez, line 220: Moving this into here as it should only run on the old version.

src/middlewares/users.ts
```diff
@@ -222,69 +193,51 @@ export async function putUser(
...
-    allAddresses = allAddresses.concat(req.body.verifiedAddresses)
-    identityKey = req.body.identityKey
-  } else if (requestVersion >= adminApiVersions[1]) {
-    const addressesAndKey = parseVerifiedAddresses(req.body.verifiedAddresses)
-    identityKey = addressesAndKey[1]
-    allAddresses = allAddresses.concat(addressesAndKey[0])
-  }
-
-  let updatedAddresses
-  let statusCode = HttpStatus.OK
```
dino-rodriguez, line 245: No need to set `OK` -- this automatically gets set in the `sendSuccess` middleware.

src/services/users.ts
```diff
@@ -1,7 +1,67 @@
...
+ * so they can be consumed by the database.
+ *
+ * @param maybeAddresses - An array of addresses ( in either the old or new format ) or undefined.
+ * @param maybeVerifiedAddresses - An array of verified addresses ( in either the old or new format ) or undefined.
+ * @param maybeIdentityKey - An identity key or undefined ( included with verified addresses ).
+ * @param requestVersion - The request version to determine how to parse the addresses.
+ *
+ * @returns A tuple of all the formatted addresses & the identity key.
+ */
+export default function parseAllAddresses(
```
dino-rodriguez, line 21: This is the helper - we were able to move all of the `undefined` checks into this function so they don't have to be repeated across `PUT` & `POST`. 

Writing this function also made it clear to me how convenient it can be to work with `maybe` constructs.

src/services/users.ts
```diff
@@ -10,7 +70,7 @@ import { ParseError, ParseErrorType } from '../utils/errors'
  *
  * @returns Array of address inforation to be consumed by insertUser.
  */
-export default function parseVerifiedAddresses(
+function parseVerifiedAddresses(
```
dino-rodriguez, line 73: This now becomes a helper.

PR #687, opened 2020-09-02
Log fatal errors when running tests

intelliot:

Currently, all logging is suppressed when tests are run. This is great because logging is too noisy.

However, `FATAL` logs should not occur when everything is going well. But if the tests fail, it is useful to see `FATAL` log output.

This PR sets the log level to `FATAL` during test execution.

### Context of Change

If Postgres is not running or the database does not exist (which can also occur if there is a conflict for Postgres's port, 5432 by default), `npm run test` would fail without any useful log output.

Setting the log level to `FATAL` addresses this. This PR also adds a suggested next step for the particular error that I encountered.

### Type of Change

Note: This is only a new feature for those who run the payid tests (`npm run test`). It has no effect on any other aspect of this project.

## Before / After

Before:
Screen Shot 2020-09-02 at 8 56 09 AM

After:
Screen Shot 2020-09-02 at 8 56 45 AM

## Test Plan

`npm run test` when Postgres is not running or the database does not exist

### DIFF
--- a/src/db/syncDatabaseSchema.ts
+++ b/src/db/syncDatabaseSchema.ts
@@ -72,7 +72,12 @@ async function executeSqlFile(
     // Close the database connection
     await client.end()
   } catch (err) {
-    logger.fatal('error running query', file, err.message)
+    logger.fatal(
+      '\nerror running query',
+      file,
+      err.message,
+      '\n\nCheck that Postgres is running and that there is no port conflict\n',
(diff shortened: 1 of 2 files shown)


dino-rodriguez (approved):
LGTM - Thanks Elliot!

austin-king (approved):
LGTM

PR #693, opened 2020-09-16
Update config.ts

VIRTUECOIN:
### Type of Change

- [ ] Documentation Updates
- [x] Release

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -49,7 +49,7 @@ const config = {
      *
      * This will be dynamically set by incoming requests if the ENV var is unset.
      */
-    domain: process.env.PAYID_DOMAIN ?? 'missing_domain',
+    domain: process.env.PAYID_DOMAIN ?? 'Virtuecoin.zil',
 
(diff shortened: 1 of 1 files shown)


VIRTUECOIN:
Was this done correctly?

hbergren:
You shouldn't need to change that variable. This should get automatically set by the first HTTP request that hits the Public API of your PayID server. If your server is actually on that domain, that should happen automatically.

VIRTUECOIN:
What ya mean

dino-rodriguez:
@VIRTUECOIN When you set up a PayID server, it will automatically get your domain from the first request made to the server, so you don't need to set this value unless you want to specify a different domain than the one your server is running.

This is also a configuration setting that should be changed locally in your server, as opposed to in this repository since it is a change that is personal to VirtueCoin.

Does that make sense?

VIRTUECOIN:
Im a newbie it kinda makes sense could you help me get this going I want to
airdrop virtuecoin also

